PR #48, opened 2023-03-01
fix(css): set height for `.Main` to ensure height fill on mobile

bkrem:
- Helps with the mobile chat window issue shown in screenshot for now, but not a final solution to this.

Screenshot 2023-03-01 at 06 49 05

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -35,6 +35,7 @@
     @include padding-x(0.4em);
     grid-area: main;
     max-width: 100vw;
+    height: 96vh;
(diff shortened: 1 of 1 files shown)


devceline (approved):
LGTM, I have an idea of how to fix this permanently, will push a "cleaner" fix tomorrow

bkrem:
> LGTM, I have an idea of how to fix this permanently, will push a "cleaner" fix tomorrow

Sounds good 💯

PR #49, opened 2023-03-01
fix(threads-height): removed height 100% to avoid pushing the empty box to the bottom

Cali93:
### Context

- Removed height 100% on threads classname to avoid pushing the next element down to the bottom
- Replaced `flex-flow: column wrap` with `flex-direction: column` to avoid issues on smaller devices (eg: iPhone SE)


bkrem (approved):

bkrem:
💯💯

devceline (approved):
LGTM

PR #50, opened 2023-03-07
Implement new message queue

devceline:
# Changes
- Implements a simplistic message queue

# Note
~~This PR works, theoretically, and will work practically (:crossed_fingers: ) once [this PR](https://github.com/WalletConnect/chat-client-js/pull/35) gets merged in~~

This PR works now that [this PR](https://github.com/WalletConnect/chat-client-js/pull/35) has been merged in. 

This feature does not work on Firefox, but does work on chrome. This feature working on Firefox is blocked by an issue in core and has to do with re-connection, outside the scope of this PR.

Showing and styling unsent messages will be done in another PR, this is purely functionality.

# Message queue description
- When `message` is called at the top level, it simply "queues" the message by populating a `ReplaySubject` ([Read More](https://rxjs.dev/api/index/class/ReplaySubject))
- One subscriber is permanently listening to the ReplaySubject, it does the following:
    - Attempt to publish message using provider
    - If successful, that's the end of the message's cycle
    - If failed, it reques the message with an updated timestamp and incremented count
- The permanent subscriber filters messages that have a count < 3
- There's an added function that retrieves unsent messages. This is at the facade level since it is not the provider's business
- The above function works by creating a temporary subscription, and since `ReplaySubject` gives new subscriber's historic data, this way we can create a sync operation that just retrieves history
- We can then mutate the data to show a status, eg: `count > 3 === "FAILED"` otherwise it is pending. 

## Showcase of `ReplaySubject` working for easy explanation
Wrote this to figure out how well it'd work for unsent messages, thought it'd help explain if unfamiliar with `ReplaySubject`.

Every `Active` console.log happens after a click, and every `Temp` console.log happens after a double click.
 
image

### DIFF
--- a/src/components/messages/ThreadWindow/index.tsx
+++ b/src/components/messages/ThreadWindow/index.tsx
@@ -61,7 +61,7 @@ const ThreadWindow: React.FC = () => {
     if (!chatClientProxy) {
       return noop
     }
-    const sub = chatClientProxy.observe('chat_message', {
+    const receivedMessageSub = chatClientProxy.observe('chat_message', {
       next: messageEvent => {
         /*
          * Ignore message events from other threads
@@ -74,6 +74,10 @@ const ThreadWindow: React.FC = () => {
 
         refreshMessages()
       }
+    })
+
+    const sentMessageSub = chatClientProxy.observe('chat_message_sent', {
+      next: refreshMessages
     })
 
     const inviteAcceptedSub = chatClientProxy.observe('chat_invite_accepted', {
@@ -97,7 +101,8 @@ const ThreadWindow: React.FC = () => {
     return () => {
       inviteAcceptedSub.unsubscribe()
       inviteRejectedSub.unsubscribe()
-      sub.unsubscribe()
+      receivedMessageSub.unsubscribe()
+      sentMessageSub.unsubscribe()
     }
   }, [chatClientProxy, refreshMessages, topic, nav, peer])
 
@@ -142,11 +147,7 @@ const ThreadWindow: React.FC = () => {
           ))}
         </AnimatePresence>
       </div>
-      <MessageBox
-        onSuccessfulSend={refreshMessages}
-        authorAccount={`eip155:1:${userPubkey ?? ''}`}
-        topic={topic}
-      />
+      <MessageBox authorAccount={`eip155:1:${userPubkey ?? ''}`} topic={topic} />
     </div>
   )
 }
--- a/src/w3iProxy/chatProviders/internalChatProvider.ts
+++ b/src/w3iProxy/chatProviders/internalChatProvider.ts
@@ -204,7 +204,15 @@ export default class InternalChatProvider implements W3iChatProvider {
       throw new Error(this.formatClientRelatedError('message'))
     }
 
-    await this.chatClient.message(params)
+    const isConnected = this.chatClient.core.relayer.provider.connection.connected
+
+    console.log({ isConnected })
+
+    try {
+      await this.chatClient.message(params)
+    } catch {
+      throw new Error('Message failed')
+    }
 
     return Promise.resolve()
   }
--- a/src/w3iProxy/chatProviders/internalChatProvider.ts
+++ b/src/w3iProxy/chatProviders/internalChatProvider.ts
@@ -21,29 +21,10 @@ export default class InternalChatProvider implements W3iChatProvider {
         return
       }
 
-      console.log(
-        'Currently connected: ',
-        this.chatClient.core.relayer.connected,
-        'Connecting: ',
-        this.chatClient.core.relayer.connecting
-      )
-
       if (!this.chatClient.core.relayer.connected) {
-        this.chatClient.ping({ topic: '' }).then(() => {
-          console.log('Pinged')
-        })
+        // Ping empty topic to trigger reconnection mechanism
+        this.chatClient.ping({ topic: '' })
       }
-
-      /*
-       * This.chatClient.core.relayer.provider.connection
-       *   .close()
-       *   .catch(() => null)
-       *   .then(() => {
-       *     console.log('Closed and opened')
-       *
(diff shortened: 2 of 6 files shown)


Cali93 (commented):
LGTM, the only issue I found was that messages were sent twice when following those steps:
1. go offline
2. send the message
3. go online

src/components/messages/ThreadWindow/index.tsx
```diff
@@ -142,11 +147,7 @@ const ThreadWindow: React.FC = () => {
           ))}
         </AnimatePresence>
       </div>
-      <MessageBox
-        onSuccessfulSend={refreshMessages}
-        authorAccount={`eip155:1:${userPubkey ?? ''}`}
-        topic={topic}
-      />
+      <MessageBox authorAccount={`eip155:1:${userPubkey ?? ''}`} topic={topic} />
```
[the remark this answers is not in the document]
devceline: We can, but I'd rather leave this for another PR since we have `eip155:1` all over the place, we can leave it for another refactor

src/w3iProxy/chatProviders/internalChatProvider.ts
```diff
@@ -15,6 +16,17 @@ export default class InternalChatProvider implements W3iChatProvider {
   public constructor(emitter: EventEmitter) {
     this.emitter = emitter
 
+    interval(2000).subscribe(() => {
+      if (!this.chatClient) {
+        return
+      }
+
+      if (!this.chatClient.core.relayer.connected) {
+        // Ping empty topic to trigger reconnection mechanism
```
bkrem, line 25: Assuming this is a workaround until https://github.com/WalletConnect/walletconnect-monorepo/pull/2108 has been released?

src/w3iProxy/w3iChatFacade.ts
```diff
@@ -11,9 +11,19 @@ import type {
 } from './chatProviders/types'
 import InternalChatProvider from './chatProviders/internalChatProvider'
 import ExternalChatProvider from './chatProviders/externalChatProvider'
+import { distinct, filter, from, ReplaySubject, throwError, timeout } from 'rxjs'
+// eslint-disable-next-line no-duplicate-imports
 import { fromEvent } from 'rxjs'
```
bkrem, line 16: Do we need the two distinct import statements here since it's both just `import`, not `import` + `import type`?

bkrem: Might as well disable this lint rule anyways since it keeps flagging `import` + `import type` which is widely used in the w3i codebase

src/w3iProxy/w3iChatFacade.ts
```diff
@@ -35,6 +47,58 @@ class W3iChatFacade implements W3iChat {
 
     const ProviderClass = this.providerMap[this.providerName]
     this.provider = new ProviderClass(this.emitter)
+
+    // Discuss expiry of messages
```
bkrem, line 51: Context here? I.e. message expiry is unclear for retries specifically or beyond that too?

src/w3iProxy/w3iChatFacade.ts
```diff
@@ -35,6 +47,58 @@ class W3iChatFacade implements W3iChat {
...
+             * Create arbitrary delay.
+             * The timestamp is updated to be the current time.
+             */
+            setTimeout(() => {
+              this.messageReplaySubject.next({
+                ...params,
+                count: params.count + 1,
+                timestamp: Date.now()
+              })
+            }, this.messageSendTimeout * 1.25)
```
bkrem, line 83: Why do we need a delay here and why the magic number `1.25` as a multiplier for what seems to be an already defined timeout value?

src/w3iProxy/w3iChatFacade.ts
```diff
@@ -35,6 +47,58 @@ class W3iChatFacade implements W3iChat {
...
+  }
+
+  private handleMessagePublishing() {
+    // Stop trying to resend the message after 3 tries.
+    this.messageReplaySubject.pipe(filter(params => params.count < 3)).subscribe(params => {
+      from(this.provider.message(params))
+        .pipe(
+          timeout({
+            first: this.messageSendTimeout,
+            with: () => throwError(() => new Error())
```
bkrem, line 63: Should this throw with a concrete error?

bkrem (approved):
Approving since you mentioned there's still iteration coming here anyways once the Core fix is published etc. 

Mainly some contextual aspects that are still unclear to me rn in the implementation

src/w3iProxy/w3iChatFacade.ts
```diff
@@ -118,11 +182,14 @@ class W3iChatFacade implements W3iChat {
...
-        ...params
-      })
+    this.messageReplaySubject.next({
+      ...params,
+      originalTimestamp: params.timestamp,
+      id: `${params.message}${params.timestamp}${params.authorAccount}`,
+      count: 0
     })
+
+    return Promise.resolve()
```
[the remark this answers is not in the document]
Cali93: What I suggested doesn't return a value, you could even use it to compare the last value with the provided params to make sure the message has been sent.

It feels hacky/coding for the linter to be happy with the current `return Promise.resolve()`

src/w3iProxy/w3iChatFacade.ts
devceline: Just wanted it to be longer than the time of the `timeout` function, `1.25` is entirely arbitrary

src/w3iProxy/w3iChatFacade.ts
devceline: Right now, we do nothing with the error so I don't see the need to, we could possibly change this if in the future `provider.request` in core threw errors in relation to connection, but for now, there's nothing to do with the errors other than catch and retry

Cali93 (approved):
LGTM as a first iteration and acknowledging that offline messages are behaving a bit unexpectedly.

PR #51, opened 2023-03-07
Refactor/external communication

devceline:
> *Prefer composition over inheritance*

# Changes
- Move from an inheritance based approach for external communicators to composition based *Communicator* modules. 
- This simplifies the code from having duplicate android/ios communicators
- Nothing changes with how communication actually happens, just moved code around


bkrem:
Could you update the PR description with any relevant context pls? 

Overall seems to be consolidating the provider pattern usage to be more DRY as we talked about a while back which is great, is there anything else to look out for during review?

devceline:
> Could you update the PR description with any relevant context pls?
> 
> Overall seems to be consolidating the provider pattern usage to be more DRY as we talked about a while back which is great, is there anything else to look out for during review?

Updated PR description

src/w3iProxy/pushProviders/internalPushProvider.ts
```diff
@@ -8,7 +8,7 @@ export default class InternalPushProvider implements W3iPushProvider {
   private readonly emitter: EventEmitter
   public providerName = 'InternalChatProvider'
 
-  public constructor(emitter: EventEmitter) {
+  public constructor(emitter: EventEmitter, _name = 'internal') {
```
bkrem, line 11: Guess I'm missing something but I'm actually not clear on the `_` prefix here, why not just do default assignment as `name = 'internal'`?

Cali93 (approved):
LGTM, feels cleaner this way 👌

src/w3iProxy/w3iPushFacade.ts
```diff
@@ -9,17 +9,15 @@ import type {
...
-import iOSPushProvider from './pushProviders/iosPushProvider'
 import InternalPushProvider from './pushProviders/internalPushProvider'
 import type { PushFacadeEvents } from './listenerTypes'
 
 class W3iPushFacade implements W3iPush {
   private readonly providerMap = {
     internal: InternalPushProvider,
     external: ExternalPushProvider,
-    android: AndroidPushProvider,
-    ios: iOSPushProvider
```
bkrem, line 22: Can we also delete the `iOSPushProvider`?

src/w3iProxy/pushProviders/externalPushProvider.ts
```diff
@@ -2,18 +2,34 @@ import type { EventEmitter } from 'events'
...
+      case 'ios':
+        this.communicator = new IOSCommunicator(this.emitter)
+        break
+      default:
+        this.communicator = new JsCommunicator(this.emitter)
+        break
+    }
   }
 
   protected async postToExternalProvider<MName extends keyof PushClientFunctions>(
```
bkrem, line 35: Should this also delegate to the `communicator` here as for `externalChatProvider`? https://github.com/WalletConnect/web3inbox/pull/51/files#diff-1f7b1977dc91a5064fee19ccb242c192b43705ceb88616ca0c87a7ad61a8091eR48

bkrem (approved):
✨

src/w3iProxy/pushProviders/internalPushProvider.ts
devceline: Simply to denote it's not used anywhere

src/w3iProxy/w3iPushFacade.ts
devceline: There is no `iOSPushProvider`

PR #52, opened 2023-03-08
Feat/improved signin flow

devceline:
# Changes

- Change identity key signing mechanism
- Add 2 new events to achieve this, a ping and a pong. `chat_signature_requested`, `chat_signature_delivered`.
- Change signature modal styling
- This helps mobile sign in flows.

### DIFF
--- a/src/pages/Login/index.tsx
+++ b/src/pages/Login/index.tsx
@@ -29,7 +29,7 @@ const Web3InboxFeatures = [
 ]
 
 const Login: React.FC = () => {
-  const { userPubkey, registeredKey, isRegistering } = useContext(W3iContext)
+  const { userPubkey, registeredKey, registerMessage } = useContext(W3iContext)
   const nav = useNavigate()
 
   useEffect(() => {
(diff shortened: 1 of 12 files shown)


src/pages/Login/index.tsx
```diff
@@ -74,7 +74,11 @@ const Login: React.FC = () => {
         </div>
         <div className="Login__actions">
           <div className="Login__actions-container">
-            {userPubkey && !registeredKey && isRegistering ? <SignatureModal /> : <Web3Button />}
+            {userPubkey && !registeredKey && registerMessage ? (
```
Cali93, line 77: It looks like there is an edge case that isn't handled.
Steps to reproduce:

1. Connect
2. Sign the message
3. Change your account
4. You are blocked on the login page

We might need a nested ternary here to display a Sign Message button if the user is connected but doesn't have registeredKey

Cali93 (commented):
Nice job, there's just an edge case we need to take care of and it'll be 👌

PR #54, opened 2023-03-10
Feat/swipe thread

boidushya:
Draft for swiping notification feature.


src/components/messages/ThreadSelector/Thread/Thread.scss
```diff
@@ -1,3 +1,85 @@
+// .Thread {
```
Cali93, line 1: Why do we have commented code ?

Please remove it if it's unused.

src/components/messages/ThreadSelector/Thread/index.tsx
```diff
@@ -1,3 +1,134 @@
+/*
```
Cali93, line 1: Same here, please remove the code instead of commenting if it's unused

src/components/notifications/AppNotifications/AppNotificationItem.tsx
```diff
@@ -1,7 +1,13 @@
...
+import { useContext, useEffect, useState } from 'react'
 import { useFormattedTime } from '../../../utils/hooks'
 import CircleIcon from '../../general/Icon/CircleIcon'
 import './AppNotifications.scss'
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
+import { LazyMotion, PanInfo, domMax, m, useAnimationControls } from 'framer-motion'
+import ClearIcon from '../../../assets/ClearIcon.png'
+import UnreadIcon from '../../../assets/UnreadIcon.png'
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
+import { AppNotificationDragContext, AppNotificationsDragContext } from './index'
```
Cali93, line 10: I think it's expecting you to import the type as such:

import type { AppNotificationsDragContext } from './index'

src/components/notifications/AppNotifications/AppNotificationItem.tsx
```diff
@@ -16,34 +22,131 @@ interface IAppNotificationProps {
...
+    if (shouldShowActions) {
+      actionControls.start('visible')
+    } else {
+      actionControls.start('hidden')
+    }
+  }
+
+  // Change current items drag state to true and all others to false
+  const handleDragStart = () => {
+    setNotificationsDrag(
```
Cali93, line 81: It might be safer to use the currentState:

```ts
setNotificationsDrag(currentNotifications =>
      currentNotifications?.map(n => {
        return {
          ...n,
          isDragged: n.id.toString() === notification.id
        }
      })
    )
```

Cali93 (commented):
Good job so far, just a few minor comments code wise.

And regarding the UI it looks nice but there's 2 issues:

- Can you please make sure the sidebar doesn't get pushed way down ?
- Can you make it so that the notification doesn't overflow the view width ?

 
Capture d’écran 2023-03-10 à 08 52 45
Capture d’écran 2023-03-10 à 08 52 37

src/components/messages/ThreadSelector/Thread/Thread.scss
boidushya: Okay! Slipped my mind that I'd commented it there, ill get rid of it

src/components/notifications/AppNotifications/AppNotificationItem.tsx
boidushya: Correct, but doing so gives another eslint issue, I'll try taking a look and hashing it out

src/components/notifications/AppNotifications/AppNotificationItem.tsx
boidushya: Ah! Perfect, ill get that goin

devceline (commented):
Looks good to me barring the comments @Cali93 left, but in desktop mode some of the badge gets cuts away, and completely goes away when hovering.

image

I disabled the following rule:

image



And got this result, which looks like everything is working in:

image

Why do we need that `overflow-x: hidden` when `min-width > 700`?

PR #60, opened 2023-03-13
Added animations for new messages sent and received

boidushya:
# Behavior
A bounce like animation added for incoming and outgoing messages

# Example
Message sent:

https://user-images.githubusercontent.com/26746725/224707332-5670bf35-5646-4ffa-8bfe-668b20086ec4.mov

Message received:

https://user-images.githubusercontent.com/26746725/224707347-85b32531-66b5-4f2d-ae6a-057607ef10e6.mov

### DIFF
--- a/src/components/messages/Message/Message.scss
+++ b/src/components/messages/Message/Message.scss
@@ -1,7 +1,51 @@
+$bounce-curve: cubic-bezier(0.34, 1.56, 0.64, 1);
+
 .Message {
+  $parent: &;
+
   display: flex;
   position: relative;
-  animation: bounce-up 0.2s cubic-bezier(0.34, 1.56, 0.64, 1) forwards;
+  &--recent {
+    animation: bounce-up 0.2s $bounce-curve forwards;
+  }
+  &__container {
+    &--first {
+      #{$parent}__from-peer #{$parent}__bubble {
+        border-radius: 16px 16px 16px 6px;
+      }
+
+      #{$parent}__from-sender #{$parent}__bubble {
+        border-radius: 16px 16px 6px 16px;
+      }
+    }
+    &--last {
+      #{$parent}__from-peer #{$parent}__bubble {
+        border-radius: 6px 16px 16px 16px;
+      }
+
+      #{$parent}__from-sender #{$parent}__bubble {
+        border-radius: 16px 6px 16px 16px;
+      }
+    }
+    &--same {
+      #{$parent}__from-peer #{$parent}__bubble {
+        border-radius: 6px 16px 16px 6px;
+      }
+
+      #{$parent}__from-sender #{$parent}__bubble {
+        border-radius: 16px 6px 6px 16px;
+      }
+    }
+    &--solo {
+      #{$parent}__from-peer #{$parent}__bubble {
+        border-radius: 16px 16px 16px 16px;
+      }
+
+      #{$parent}__from-sender #{$parent}__bubble {
+        border-radius: 16px 16px 16px 16px;
+      }
+    }
+  }
 
   &__from-peer {
     justify-content: flex-start;
@@ -15,20 +59,19 @@
     max-width: 50%;
     overflow-wrap: break-word;
     padding: 0.4em 0.8em;
-    margin-top: 0.5em;
+    will-change: border-radius;
+    transition: border-radius ease 0.2s;
   }
 
   &__from-peer &__bubble {
     background: var(--sender-bubble-bg);
     box-shadow: var(--sender-box-shadow);
-    border-radius: 16px 16px 16px 6px;
     color: var(--fg-color-1);
   }
 
   &__from-sender &__bubble {
     box-shadow: inset 1px 1px 4px #99cbff, inset -1px -1px 4px #1464b8;
     background: #3396ff;
-    border-radius: 16px 16px 6px 16px;
     color: #ffffff;
   }
 }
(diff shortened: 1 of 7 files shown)


boidushya:
Added message box states based on previous and next message

### Preview
image

src/components/messages/Message/Message.scss
```diff
@@ -13,20 +59,30 @@
...
 
   &__from-sender &__bubble {
     box-shadow: inset 1px 1px 4px #99cbff, inset -1px -1px 4px #1464b8;
     background: #3396ff;
-    border-radius: 16px 16px 6px 16px;
     color: #ffffff;
   }
 }
+
+@keyframes bounce-up {
```
Cali93, line 79: Sorry for the lack of context on the issues but the animation should be a scaling animation.
You can use [this](https://github.com/WalletConnect/minisite-web3inbox/blob/e5aa0dc9bbc9f5803b53c60052bed6ecf1a6d4f4/src/lib/components/messages/SenderBubble.tsx#L22) as a reference

src/components/messages/Message/Message.scss
```diff
@@ -1,5 +1,51 @@
...
 .Message {
+  $parent: &;
+
   display: flex;
+  position: relative;
+  &--recent {
+    animation: bounce-up 0.2s $bounce-curve forwards;
+  }
+  &__container {
+    &--first {
```
Cali93, line 12: This looks a bit overkill and doesn't match the behavior designed in Figma, please take a look at [this frame](https://www.figma.com/file/P70QpTDpsgwdPJ2jJmVXxg/web3inbox-standalone?node-id=85%3A6093&t=PnaTzlxGXGYowXT2-4).
You'll notice that the gap between "stacked" messages is smaller and the top right and bottom right border radius are 6px.

src/components/messages/Message/Message.scss
```diff
@@ -1,5 +1,51 @@
+$bounce-curve: cubic-bezier(0.34, 1.56, 0.64, 1);
+
 .Message {
+  $parent: &;
```
devceline, line 4: Can you give more context as to what's going on here?

Cali93 (commented):
Let's try to match the Figma as close as we can and it'll be good to go.

src/components/messages/Message/MessageItem.tsx
```diff
@@ -9,8 +9,19 @@ interface IMessageItemProps {
...
+
+type ILocalMessageType = 'first' | 'last' | 'same' | 'solo'
+export const MessageItem: React.FC<IMessageItemProps> = ({
+  messages,
+  message,
+  index,
+  peer,
+  previousMessageAccount,
+  nextMessageAccount
+}) => {
```
devceline, line 24: The responsibility of knowing the next and previous message, and changing styling accordingly should be that of a parent container. I'm not saying this is a total blocker, but if we keep this in like this, it will be refactored at some point when new requirements come up.

src/components/messages/ThreadWindow/ThreadWindow.scss
```diff
@@ -16,6 +16,7 @@
   }
 
   &__messages {
+    height: 100%;
```
devceline, line 19: I thought we were avoiding more `height: 100%`, why do we need this here?

src/components/messages/Message/Message.scss
boidushya: Since using `&` would refer to its local parent instead of the root element, I am storing it in a variable called `$parent` to use it inside bubble classes to define the border radius

src/components/messages/Message/MessageItem.tsx
boidushya: Sounds good, I'll take a look at this then. 
The reason why I added this was to maintain the classnames for the children. 
Technically we could group them in another div based on the parents and style them accordingly, but that was leading to a lot of regrouping happening amongst divs since every new message would change the structure of divs and it didnt feel optimal at the time.

src/components/messages/ThreadWindow/ThreadWindow.scss
boidushya: Must've been a stray I accidentally left in, thanks for pointing this out!

boidushya:
### Title
Revamped animations for messages. 

### Example

https://user-images.githubusercontent.com/26746725/225099609-ccc2562e-9707-4adb-9dee-c4ab52dd3670.mov

src/components/messages/Message/Message.scss
```diff
@@ -13,20 +46,40 @@
...
   }
 
   &__from-sender &__bubble {
     box-shadow: inset 1px 1px 4px #99cbff, inset -1px -1px 4px #1464b8;
     background: #3396ff;
-    border-radius: 16px 16px 6px 16px;
     color: #ffffff;
   }
 }
+
```
Cali93, line 65: Please remove commented code

Cali93 (approved):
LGTM, good job! Don't forget to remove any commented code.

src/components/messages/Message/Message.scss
boidushya: oops, right, fixing right away!

src/components/messages/Message/MessageItem.tsx
boidushya: Noting this down here for later:

`Conversation` component, which would encompass `ConversationBeginning` and all `MessageItems`, this container would structure message. For example, when mapping messages to `MessageItem` components, it would check previous and next messages in the map by using the index provided in the map

PR #62, opened 2023-03-14
Add SearchSuggestion component

devceline:
# Changes
- Add SearchSuggestion component
- Use it in new chat
- Make it work in mobile and desktop
- Using `thegraph` for its ens domain subgraph

# Note
Although it does use graphql, I did not feel it appropriate to add a graphql package or spend time on a util when this is our only graphql use case, therefore I went with a hardcoded query string.

### DIFF
--- a/src/components/messages/NewChat/index.tsx
+++ b/src/components/messages/NewChat/index.tsx
@@ -95,8 +95,6 @@ const NewChat: React.FC = () => {
     [query, isInviting]
   )
 
-  console.log('Searchhh')
-
   return (
     <Fragment>
       <div className="NewChat">
--- a/src/components/messages/NewChat/index.tsx
+++ b/src/components/messages/NewChat/index.tsx
@@ -1,5 +1,5 @@
 import { fetchEnsAddress } from '@wagmi/core'
-import React, { Fragment, useCallback, useContext, useMemo, useState } from 'react'
+import React, { Fragment, useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { toast } from 'react-toastify'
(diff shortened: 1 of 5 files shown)


Cali93 (commented):
LGTM, minor comment on the gql query that could be formatted using `` to improve readability.

src/components/messages/NewChat/index.tsx
```diff
@@ -94,12 +95,15 @@ const NewChat: React.FC = () => {
     [query, isInviting]
   )
 
+  console.log('Searchhh')
```
boidushya, line 98: Lets get rid of this before we forget :)

boidushya (commented):
LGTM, except what @Cali93 suggested about using template literals and the stray `console.log`

src/components/messages/NewChat/index.tsx
devceline: Good catch :sweat:

devceline (commented):
Found a bug due to the old debouncing mechanism, so I updated it

boidushya (approved):
LGTM, the search is a _wee_ bit slow at the time but i think that can be fixed later on 🙌

Also for future reference maybe we could also make sure if the exact typed ENS address exists in results, we show that at the top? 
image

devceline:
> LGTM, the search is a _wee_ bit slow at the time but i think that can be fixed later on raised_hands
> 
> Also for future reference maybe we could also make sure if the exact typed ENS address exists in results, we show that at the top? image

That sounds interesting, we could do it with one extra call I think

Issue #59, opened 2023-03-13
Fix issues spotted in Design Review


devceline:
Issues spotted in design review meeting (excluding the more involved issues):

- [x] new chat icon too large
- [x] change placeholder in new chat space to slash
- [x] When inviting, it would go to the chat
- [x] messages start at the bottom and chat invite pending should be right above the input
- [x] messages should animate on appearance (mirrored for sender)
- [x] join messages that were sent less than header show time
- [x] MessageBox focus state (blue border)
- [x] Max width for the thread selector: 280px
- [x] Independent scroll for thread selector
- [x] Checkbox styling
- [x] Check icon in accepted should be white
- [x] Send invite disabled color is wrong
- [x] Send button color is wrong
- [x] Disabled button style incorrect in light mode
- [x] Checkbox is incorrect in lightmode
- [x] Hide invites when searching
- [x] Max message width in mobile should be 80%
- [x] too much space between input(MessageBox) and bottom
- [x] message timestamps should be further to the right
- [x] Transition animation between sign in and dapp
- [x] Include the notification bullet point on the hover area

cc @Cali93 @boidushya

chadyj:
Also it looks like the Settings don't match Figma.
https://www.figma.com/file/P70QpTDpsgwdPJ2jJmVXxg/web3inbox-standalone?node-id=251%3A16017&t=4Ls93Y9P9Wu6KIDv-1

PR #64, opened 2023-03-15
Fix/styling

boidushya:
## Title

Several Styling fixes as per #59 

### new chat icon too large
image

### change placeholder in new chat space to slash
image

### MessageBox focus state (blue border)
image

### Max width for the thread selector: 280px
(Changed in code, also modified `min-width` which should probably be adjusted later)

### Max message width in mobile should be 80%
image

### DIFF
--- a/src/components/messages/Message/Message.scss
+++ b/src/components/messages/Message/Message.scss
@@ -48,6 +48,9 @@
     padding: 0.4em 0.8em;
     will-change: border-radius;
     transition: border-radius ease 0.2s;
+    @media only screen and (max-width: 700px) {
+      max-width: 80%;
+    }
   }
 
   &__from-peer &__bubble {
--- a/src/components/messages/Message/Message.scss
+++ b/src/components/messages/Message/Message.scss
@@ -43,14 +43,13 @@
   }
 
   &__bubble {
-    max-width: 50%;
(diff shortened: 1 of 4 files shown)


src/components/messages/Message/Message.scss
```diff
@@ -48,6 +48,9 @@
     padding: 0.4em 0.8em;
     will-change: border-radius;
     transition: border-radius ease 0.2s;
+    @media only screen and (max-width: 700px) {
+      max-width: 80%;
+    }
```
devceline, line 53: We like to go with mobile first, so if it should be `max-width: 80%` on mobile, then the default should be `max-width: 80%` and then have a media query `(min-width: 700px)`.

devceline:
```suggestion
    max-width: 80%;
    @media only screen and (min-width: 700px) {
        max-width: unset;
    }
```

devceline (commented):
Looks great, just one comment

src/components/messages/Message/Message.scss
boidushya: Ah alright! Sounds great :)
Will keep that in mind in future as well 🙌

Cali93 (approved):
LGTM 👍

PR #65, opened 2023-03-15
Feat/limit invite to keyserver

devceline:
# Changes
- Provide reasons for send invite button being disabled
- Disable button if account does not exist on keyserver


boidushya (approved):
LGTM
Do take a look at @Cali93 's suggestion btw :))

PR #70, opened 2023-03-22
Fix/styling

boidushya:
## Title

More Styling fixes as per #59 

### Page Transitions between sign in and dApp

https://user-images.githubusercontent.com/26746725/227024979-cd8cc6bc-c05c-414b-a89b-4d096a364359.mov

### Checkmark color in light mode

image
image

### Timestamp position on mobile devices

image

### Lack of gap between message box and threads
image


src/components/general/Icon/CheckIcon.tsx
```diff
@@ -1,18 +1,13 @@
...
-  const themeColors = useColorModeValue(mode)
-
   return (
     <svg width="14" height="12" viewBox="0 0 14 12" fill="none" xmlns="http://www.w3.org/2000/svg">
       <path
         fillRule="evenodd"
         clipRule="evenodd"
         d="M12.6533 0.12431C13.0032 0.345935 13.1072 0.809277 12.8856 1.15921L6.31919 11.5272C5.9518 12.1073 5.12246 12.1531 4.6935 11.6169L1.16242 7.20301C0.903666 6.87956 0.956107 6.40759 1.27955 6.14884C1.603 5.89008 2.07497 5.94252 2.33373 6.26597L5.20805 9.85887C5.31529 9.99292 5.52263 9.98149 5.61448 9.83646L11.6184 0.356635C11.84 0.00669932 12.3033 -0.0973162 12.6533 0.12431Z"
-        fill={themeColors['--fg-color-1']}
+        fill={`currentColor`}
```
Cali93, line 10: Can you give context about `currentColor` ?

Cali93 (approved):
LGTM 🚀

src/components/general/Icon/CheckIcon.tsx
boidushya: Sure, for for the checkmark color issue on light mode, previously we were using a context and switching the color when the theme was changed (except for this icon specifically, which is why the issue came to '_light_') But upon closer inspection I found that when the context was being switched, all the parent components of which the CheckIcon was a part of were changing their color to what we wanted the checkmark to be. Using `currentColor` we eliminate one performance intensive state and let CSS do it for us by referencing the closest parents' color, hence `currentColor`.


### TL;DR
I fixed the checkmark color issue on light mode by using currentColor to reference the closest parent's color, which eliminated the need for a context and improved performance.

src/components/messages/PeerAndMessage/PeerAndMessage.scss
```diff
@@ -16,6 +16,12 @@
     max-width: 15em;
   }
 
+  &__text {
+    @media screen and (max-width: 700px) {
+      max-width: none;
+    }
+  }
+
```
devceline, line 24: The preference is mobile first, can you please make this so that the default is max width is none and that the media selector is for min width instead?

devceline (approved):

src/components/general/Icon/CheckIcon.tsx
Cali93: Thanks! Linking what you sent me here for [more context ](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword).
Can you please lowercase it ?

src/components/messages/PeerAndMessage/PeerAndMessage.scss
boidushya: Sure thing, it keeps slipping my mind, ill go ahead and do that!

PR #71, opened 2023-03-23
Add authenticated redirection

devceline:
# Changes
- Login page becomes a spinner when `chatProvider !== "internal"`
- Login page now redirects to intended page when authenticated. Eg going to `/notifications` when authenticated will actually go to `/notifications` instead of `/messages`


Cali93 (approved):
LGTM

PR #77, opened 2023-03-30
Experiment/sync chat

devceline:
# Changes
- Sync `chatSentInvites` and `chatThreads` by listening to their store updates
- Hide `chatSentInvite` that already has a thread, to account for the 1-2 second delay of sync updates causing both the thread and sent invite to show up
- Conform to new spec for chat client since these changes required a package bump

# How was this tested
- Open web3inbox on brave and firefox, 
- send invites and observe the fact they're syncing on the invite being sent and accepted. 
- Send invites and observe that accepting them does not cause discrepancies. 

# Notes
These changes do add a small delay for invites etc, looking into how we can speed these up by possibly not waiting for the `syncSet` request to succeed before resolving the `set` function

### DIFF
--- a/src/contexts/W3iContext/index.tsx
+++ b/src/contexts/W3iContext/index.tsx
@@ -44,6 +44,63 @@ const W3iContextProvider: React.FC<W3iContextProviderProps> = ({ children }) => 
     pushProviderQuery ? (pushProviderQuery as Web3InboxProxy['pushProvider']) : 'internal'
   )
 
+  const refreshChatState = useCallback(() => {
+    if (!chatClient || !userPubkey) {
+      return
+    }
+
+    chatClient
+      .getReceivedInvites({ account: `eip155:1:${userPubkey}` })
+      .then(invite => setInvites(Array.from(invite.values())))
+    chatClient
+      .getSentInvites({ account: `eip155:1:${userPubkey}` })
+      .then(invite => setSentInvites(Array.from(invite.values())))
+    chatClient
+      .getThreads({ account: `eip155:1:${userPubkey}` })
+      .then(invite => setThreads(Array.from(invite.values())))
+  }, [chatClient, userPubkey, setThreads, setInvites])
+
+  useEffect(() => {
+    if (!chatClient) {
+      return noop
+    }
+
+    const inviteSub = chatClient.observe('chat_invite', {
+      next: () => {
+        refreshChatState()
+      }
+    })
+
+    const signatureSub = chatClient.observe('chat_signature_requested', {
+      next: ({ message }) => {
+        setRegisterMessage(message)
+      }
+    })
+
+    const inviteSentSub = chatClient.observe('chat_invite_sent', { next: refreshChatState })
+    const pingSub = chatClient.observe('chat_ping', {
+      next: () => {
+        console.log('Got a ping!')
+        refreshChatState()
+      }
+    })
+    const chatMessageSentSub = chatClient.observe('chat_message_sent', { next: refreshChatState })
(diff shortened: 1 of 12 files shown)


boidushya (approved):
LGTM

Tested it out on Arc/Brave, worked perfectly!

bkrem:
> Open web3inbox on brave and firefox,

Why these two specifically?

bkrem:
Seeing not initialized validation failures after the initial connect step:

Screenshot 2023-04-14 at 14 58 30

Based on the stack trace it looks like it's calling `.getAll()` before the client is initialized.

devceline:
> Seeing not initialized validation failures after the initial connect step:
> 
> Screenshot 2023-04-14 at 14 58 30
> 
> Based on the stack trace it looks like it's calling `.getAll()` before the client is initialized.
> 
> Also can't seem to proceed after that, I get the msg to sign but signing does nothing in w3i

Ironed out the call for getThreads pre-init and cleared up that you need to sign twice.

bkrem:
Looks like redeploy failed on types:

```
src/App.tsx(93,33): error TS2322: Type '{ purpose: string; message: string; }' is not assignable to type 'IntrinsicAttributes & { message: string; }'.
--
15:29:42.602 | Property 'purpose' does not exist on type 'IntrinsicAttributes & { message: string; }'.


```

src/contexts/W3iContext/index.tsx
```diff
@@ -153,13 +125,20 @@ const W3iContextProvider: React.FC<W3iContextProviderProps> = ({ children }) =>
     })
 
     const inviteSentSub = chatClient.observe('chat_invite_sent', { next: refreshChatState })
+    const pingSub = chatClient.observe('chat_ping', {
+      next: () => {
+        console.log('Got a ping!')
+        refreshChatState()
```
Cali93, line 131: Can you provide a bit more context on those bits ? Why do we call refreshChatState on chat_ping ?

src/contexts/W3iContext/index.tsx
```diff
@@ -169,6 +148,22 @@ const W3iContextProvider: React.FC<W3iContextProviderProps> = ({ children }) =>
     }
   }, [chatClient, refreshChatState])
 
+  useEffect(() => {
+    const handleRegistration = async () => {
+      if (chatClient && userPubkey) {
+        try {
+          const registeredKeyRes = await chatClient.register({ account: `eip155:1:${userPubkey}` })
```
Cali93, line 155: Can we create a follow up issue to replace the hardcoded chainId with the current chainId ?

Cali93 (approved):
LGTM just some minor questions around `refreshChatState()` which seems redundant

src/contexts/W3iContext/index.tsx
devceline: The sync store updates are communicated via chat pings, although this is a bit hacky, it gets the job done and we update state on all other chat events anyway

src/contexts/W3iContext/index.tsx
devceline: We definitely need to, will do

devceline:
> Looks like redeploy failed on types:
> 
> ```
> src/App.tsx(93,33): error TS2322: Type '{ purpose: string; message: string; }' is not assignable to type 'IntrinsicAttributes & { message: string; }'.
> --
> 15:29:42.602 | Property 'purpose' does not exist on type 'IntrinsicAttributes & { message: string; }'.
> ```

Fixed the issue, please take a look whenever you can :pray:

bkrem (approved):

Issue #81, opened 2023-04-07
Not easy to obtain wallet address w/o wallet app


chris13524:
When in the messaging view of Web3Inbox, there is no easy way to obtain the wallet address, without Disconnecting or using your wallet app. This is particularly a blocker for Web3Account where there is no wallet app.

![image](https://user-images.githubusercontent.com/1979423/230620485-53c518fc-b11e-4fb6-9488-17018d404a04.png)

Requesting to either:
- add a copy button next to the address, or
- use the standard Web3Modal popup which already has a copy address button in it

https://walletconnect.slack.com/archives/C044SKFKELR/p1680809723151289?thread_ts=1680806893.688679&cid=C044SKFKELR

chris13524:
Actually, you can already click the address to copy it to your clipboard, but this isn't obvious from the UI. Keeping this open because we still need some UI updates to make this user-friendly.

devceline:
Added a copy button to the address to make it more obvious

PR #82, opened 2023-04-14
Fix/modal copy

boidushya:
## Title

Signature Modal styles fixed. 
Same with avatar dropdown, fixing #81 

### Avatar Dropdown

| Mode | Before | After |
| ------------- | ------------- | ------------- |
| Hover | image | image |
| No-Hover | image | image |

_(Also comes with its own toast on successful copy or error toast if not)_
image

### Signature Modal

| Before | After |
| -- | -- |
| image | image |

### Chat Invite Header Alignment

| Before | After |
| -- | -- |
| image | image |

### Mute Button Alignment/Styles

| Before | After |
| -- | -- |
| image | image |

### DIFF
--- a/src/components/account/Avatar/Avatar.scss
+++ b/src/components/account/Avatar/Avatar.scss
@@ -25,7 +25,7 @@
     top: 0;
     bottom: inherit;
     margin-top: 0.5em;
-    width: 9em;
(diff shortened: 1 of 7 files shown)


devceline (approved):
LGTM

PR #102, opened 2023-04-26
Allow for managing what UIs are available

### LINKED ISSUE #101: [Feat] Allow configuring UI pages to display
**Describe the solution you'd like**
- Pilot partners such as Spot are looking to leverage web3inbox SDK to bootstrap Push UI, without involving additional Chat functionality.
- web3Inbox should allow for configuring which pages to show, allowing for partners and users of web3inbox to only display functionality they support.

**Suggested implementation**
- Allow for passing UI params per page to configure pages, which are all enabled by default: `chatEnabled`, `settingsEnabled`, `pushEnabled`
- When a page is configured to be disabled, it won't display in the sidebar and won't be able to be navigated to. 
- The default page is now chosen in this order: 
  1. if chat enabled: `/messages`
  2. if push enabled: `/notifications`
  3. if settings enabled: `/settings`
  4. if none: `/login`

**Additional context**
- See thread: https://walletconnect.slack.com/archives/C03RWGSBCUU/p1682433835879819

devceline:
# Description

Allow to configure UIs by setting relevant flags in params: `chatEnabled`, `pushEnabled` and `settingsEnabled`.  They are enabled by default.

If only one is enabled, the sidebar is hidden.

# Type of change

# Fixes/Resolves 

Fixes #101 

# Checklist:

# Additional Information

**Note**: Even if `chatEnabled=false`, a `chatProvider=external` must still be provided as the chat proxy is what handles accounts, for now.

### DIFF
--- a/src/contexts/W3iContext/index.tsx
+++ b/src/contexts/W3iContext/index.tsx
@@ -50,6 +50,26 @@ const W3iContextProvider: React.FC<W3iContextProviderProps> = ({ children }) => 
     wagmiDisconnect()
   }, [wagmiDisconnect])
 
+  const pushEnabledQuery = query.get('pushEnabled')
+  const chatEnabledQuery = query.get('chatEnabled')
+  const settingsEnabledQuery = query.get('settingsEnabled')
+
+  const [uiEnabled] = useState({
+    push: pushEnabledQuery ? JSON.parse(pushEnabledQuery) : true,
+    settings: settingsEnabledQuery ? JSON.parse(settingsEnabledQuery) : true,
+    chat: chatEnabledQuery ? JSON.parse(chatEnabledQuery) : true,
+    sidebar: false
+  })
+
+  const totalPagesEnabled = Object.values(uiEnabled).reduce<number>(
+    (pagesAvailable, pageEnabled) => (pageEnabled ? pagesAvailable + 1 : pagesAvailable),
+    0
+  )
+
+  if (totalPagesEnabled > 1) {
+    uiEnabled.sidebar = true
+  }
+
   useEffect(() => {
     const account = new URLSearchParams(search).get('account')
 
@@ -75,7 +95,7 @@ const W3iContextProvider: React.FC<W3iContextProviderProps> = ({ children }) => 
       return
     }
 
-    const w3iProxy = new Web3InboxProxy(chatProvider, pushProvider, projectId, relayUrl)
+    const w3iProxy = new Web3InboxProxy(chatProvider, pushProvider, projectId, relayUrl, uiEnabled)
     w3iProxy
       .init()
       .then(() => setChatClient(w3iProxy.chat))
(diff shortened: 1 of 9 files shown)


src/contexts/W3iContext/index.tsx
```diff
@@ -50,6 +50,26 @@ const W3iContextProvider: React.FC<W3iContextProviderProps> = ({ children }) =>
...
 
+  const pushEnabledQuery = query.get('pushEnabled')
+  const chatEnabledQuery = query.get('chatEnabled')
+  const settingsEnabledQuery = query.get('settingsEnabled')
+
+  const [uiEnabled] = useState({
+    push: pushEnabledQuery ? JSON.parse(pushEnabledQuery) : true,
+    settings: settingsEnabledQuery ? JSON.parse(settingsEnabledQuery) : true,
+    chat: chatEnabledQuery ? JSON.parse(chatEnabledQuery) : true,
+    sidebar: false
```
Cali93, line 61: Why is sidebar set to false by default ?

src/w3iProxy/index.ts
```diff
@@ -59,24 +63,20 @@ class Web3InboxProxy {
       projectId: this.projectId
     })
 
-    console.log('this.chatProvider', this.chatProvider)
-    console.log('this.pushProvider', this.pushProvider)
-
+    // Not disabling chatClient as it manages the account
     if (this.chatProvider === 'internal') {
```
Cali93, line 67: Don't we need the additional check as for pushClient ?
```suggestion
    if (this.chatProvider === 'internal' && this.uiEnabled.chat) {
```

Cali93 (approved):
Awesome 🚀

src/contexts/W3iContext/index.tsx
bkrem: Looks like the state is being re-evaluated below based on what was enabled:
https://github.com/WalletConnect/web3inbox/pull/102/files#diff-8837a0bdc83cb8d1f029e0555ca8879e97da5fbd68f45435f58680c167d7b3f2R70

src/w3iProxy/index.ts
bkrem: Same Q 👍

bkrem (approved):
Turbo 🔥 great to see there's a router here now too.

All worked as expected while testing, but one concern I did have is that once the sidebar goes you also lose access to the profile options behind the icon:
Screenshot 2023-04-26 at 16 19 29

Maybe not prio for w3i SDK usage but may be something we need to account for eventually.

src/w3iProxy/index.ts
devceline: I guess the comment above isn't clear enough but since chat manages account, we can't not init it even if chat ui is disabled, this will be changed when/if we make a separate auth provider

devceline:
> Turbo fire great to see there's a router here now too.
> 
> All worked as expected while testing, but one concern I did have is that once the sidebar goes you also lose access to the profile options behind the icon: Screenshot 2023-04-26 at 16 19 29
> 
> Maybe not prio for w3i SDK usage but may be something we need to account for eventually.

The case where someone manually disabled web3inbox's pages in desktop mode is beyond rare, but can definitely look into accounting for it in the future

Issue #105, opened 2023-05-01
[ Epic ] Web3Inbox milestone 3


devceline:
# Depends on
- [x] #116 
# React  Native SDK
- [x] https://github.com/WalletConnect/web3inbox/issues/126

# Push Features
- [x] #97 
- [x] #107  
- [x] Finalize `wc_pushSubscribe`
  - [x] JS
  - [x] Kotlin
  - [x] Swift

# Ready for verified dapps
- [ ] Support did documents in chat client
  - [ ] https://github.com/WalletConnect/chat-client-js/issues/50 (JS)
  - [ ] Kotlin
  - [ ] Swift
- [ ] Design unified UI
- [ ] Implement unified UI

# Nice to have
- [x] Create settings controller in web3inbox
- [x] Create auth controller in web3inbox
- [ ] Add communication with w3i-sdk settings controller
  - [ ] React Native
  - [ ] Swift
  - [ ] iOS
- [x] #106
- [ ] #113

PR #121, opened 2023-05-08
feat: integrate `pushWalletClient.subscribe` functionality + types

bkrem:
# Description

- Integrates `push-client@0.8.1`, which includes the `pushWalletClient.subscribe()` method and `push_subscription` event
- Ensures AppSelector and `Subscribe` button reflects `activeSubscriptions` state
- Removes monkey-patched type for `pushWalletClient.subscribe()`

# Type of change

Please put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

# Fixes/Resolves (Optional)

- Part of #105 

# Checklist:

- Next step is adding `update()` support:
	- Implementation: https://github.com/WalletConnect/push-client-js/tree/feat/push-update
	- Depends on https://github.com/WalletConnect/walletconnect-docs/pull/660 being merged before moving ahead with release + web3inbox
	- Will be integrated into #118


src/components/notifications/AppExplorer/AppCard/index.tsx
```diff
@@ -90,8 +94,12 @@ const AppCard: React.FC<AppCardProps> = ({ name, description, logo, bgColor, url
...
         <div className="AppCard__body__description">{description}</div>
         <div className="AppCard__body__url">{url}</div>
-        <Button className="AppCard__body__subscribe" onClick={async e => handleSubscription(e)}>
-          Subscribe
+        <Button
+          className="AppCard__body__subscribe"
+          onClick={async e => handleSubscription(e)}
+          disabled={isSubscribed}
+        >
+          {isSubscribed ? 'Subscribed' : 'Subscribe'}
```
devceline, line 102: Why do we show subscribed app cards in the first place?

src/w3iProxy/index.ts
```diff
@@ -78,6 +78,7 @@ class Web3InboxProxy {
 
     if (this.pushProvider === 'internal' && this.uiEnabled.push) {
       this.pushClient = await PushWalletClient.init({
+        logger: 'info',
```
devceline, line 81: Do we want the logger to remain in info mode or was this only for debugging?

devceline (approved):
This is very very very very cool to see in action. Amazing work :pray: I might follow up with a PR that shows a popup informing the user they need sign, using the `<SignatureModal>` for continuity.

Cali93 (approved):
LGTM 👌

src/components/notifications/AppExplorer/AppCard/index.tsx
Cali93: Wouldn't we have an unsubscribe button if we're already subscribed ?

devceline: Unsubscribing is done at the level of the subscription, through the `AppSelector`, which I'm okay with. But in an "Explore" view, there shouldn't be things you've already explored imo

Cali93: I think it might be more convenient for the end users if they can also filter the explorer apps based on subscribed/unsubscribed (which would default to all unsubscribed apps) but that's just my opinion.
Going through the AppSelector list might be a little tedious if they have a lot of subscriptions.

bkrem: > Why do we show subscribed app cards in the first place?

This was mainly me for simple state debugging, agreed that it doesn't really make sense for end users thanks for pointing out.

I've tweaked it to filter down the Explore view to exclude any apps present in `activeSubscriptions` instead

bkrem:
>  I might follow up with a PR that shows a popup informing the user they need sign, using the <SignatureModal> for continuity.

Yeah I'm actually not too sure why we're seeing that additional sign step if it should be pulling from the shared `identityKeys` class 🤔 Need to debug this more when I'm back on my laptop.

PR #159, opened 2023-06-05
Add ready checks to hooks

devceline:
# Description

- Add ready checks to prevent throwing errors on hook init
- Resolves issues where logging in -> subscribing -> logging out breaks app

# Type of change

# Checklist:

### DIFF
--- a/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
+++ b/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
@@ -8,6 +8,7 @@ export const useW3iProxy = () => {
   const projectId = import.meta.env.VITE_PROJECT_ID
 
   const { uiEnabled } = useUiState()
+  const [ready, setReady] = useState(false)
   const { chatProvider, pushProvider, authProvider } = useProviderQueries()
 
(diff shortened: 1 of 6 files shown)


src/contexts/W3iContext/hooks/w3iProxyHooks.ts
```diff
@@ -22,8 +23,8 @@ export const useW3iProxy = () => {
   )
 
   useEffect(() => {
-    w3iProxy.init()
-  }, [w3iProxy])
+    w3iProxy.init().then(() => setReady(true))
```
bkrem, line 26: Could you make `w3iProxy.initialized` a public property and use that instead or do we want a separate piece of state to signal this?

bkrem (approved):

src/contexts/W3iContext/hooks/w3iProxyHooks.ts
devceline: Web3inbox initialized is made to true at the start of the init call, to prevent multiple inits (IE from a useEffect rerunning thanks to React strict mode) from recreating clients.

This ready flag is guaranteed to only be true after initialization is fully complete

PR #169, opened 2023-07-06
Add push sync support for w3i

devceline:
# Description

- Add support for sync in push in web3inbox
- Support sync singing for push

# Type of change

# Checklist:

### DIFF
--- a/src/pages/Login/SignatureModal/index.tsx
+++ b/src/pages/Login/SignatureModal/index.tsx
@@ -45,7 +45,6 @@ export const SignatureModal: React.FC<{
           )
           break
         case 'push':
-          console.log('Sig', 'Sending push signature')
           window.web3inbox.push.postMessage(
             formatJsonRpcRequest('push_signature_delivered', { signature })
           )
--- a/src/pages/Login/SignatureModal/index.tsx
+++ b/src/pages/Login/SignatureModal/index.tsx
@@ -10,8 +10,11 @@ import Spinner from '../../../components/general/Spinner'
 import CrossIcon from '../../../components/general/Icon/CrossIcon'
 import W3iContext from '../../../contexts/W3iContext/context'
 
-export const SignatureModal: React.FC<{ message: string }> = ({ message }) => {
-  const { disconnect, userPubkey } = useContext(W3iContext)
+export const SignatureModal: React.FC<{
+  message: string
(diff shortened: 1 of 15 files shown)


src/w3iProxy/pushProviders/internalPushProvider.ts
```diff
@@ -32,16 +38,52 @@ export default class InternalPushProvider implements W3iPushProvider {
...
+      throw new Error(this.formatClientRelatedError('approve'))
+    }
+
+    return this.pushClient.enableSync({
+      ...params,
+      onSign: async message => {
+        this.emitter.emit('push_signature_requested', { message })
+
+        return new Promise(resolve => {
+          setInterval(() => {
```
bkrem, line 68: Looks like no `clearInterval` being called for this before `resolve`. My suspicion is this will keep running beyond the `resolve` without it.

bkrem: More generally: why do we have to poll here, are we missing an event or some other affordance to detect something more cleanly in syncClient?

src/pages/Login/SignatureModal/index.tsx
```diff
@@ -35,11 +38,23 @@ export const SignatureModal: React.FC<{ message: string }> = ({ message }) => {
...
-        formatJsonRpcRequest('chat_signature_delivered', { signature })
-      )
+      switch (sender) {
+        case 'chat':
+          window.web3inbox.chat.postMessage(
+            formatJsonRpcRequest('chat_signature_delivered', { signature })
+          )
+          break
+        case 'push':
+          console.log('Sig', 'Sending push signature')
```
bkrem, line 48: Still wanted?

```suggestion

```

bkrem (approved):

src/w3iProxy/pushProviders/internalPushProvider.ts
devceline: It would add more complexity, due to the fact that `chatClient.register` and `pushClient.enableSync` are fired after one another, emitting both `chat_signature_requested` and `push_signature_requested`, so the other solution is to listen to multiple events and orchestrate signing.

I went with the simpler approach, good call on `clearInterval` though. Slipped my mind

PR #176, opened 2023-07-18
Feat/widget subscribe

devceline:
# Description

- Add widget subscribe page
- Add `dappHooks` to remove responsibility of `dappContext` from `uiHooks`
- Add relevant styling to header
- Move all modals out of `App`, and into their own component. (Reduces clutter and allows displaying modals on non `App` pages like the subscribe page)
- Add redirection to `/widget/subscribe` when necessary

# Type of change

# Checklist:

### DIFF
--- a/src/contexts/W3iContext/hooks/pushHooks.ts
+++ b/src/contexts/W3iContext/hooks/pushHooks.ts
@@ -1,5 +1,6 @@
 import type { PushClientTypes } from '@walletconnect/push-client'
 import { useCallback, useEffect, useState } from 'react'
+import { useLocation } from 'react-router-dom'
 import { noop } from 'rxjs'
 import { subscribeModalService } from '../../../utils/store'
 import type { W3iPushClient } from '../../../w3iProxy'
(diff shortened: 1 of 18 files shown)


src/contexts/W3iContext/hooks/pushHooks.ts
```diff
@@ -12,6 +13,8 @@ export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean) => {
     PushClientTypes.PushSubscription[]
   >([])
 
+  const { pathname } = useLocation()
```
bkrem, line 16: Needed? Don't see this being used in the diffs but maybe I'm missing something

bkrem (approved):
Nice, liking the reorg work that happened here too ✨

src/contexts/W3iContext/hooks/pushHooks.ts
devceline: it's used to allow in the dependency array in the below `useEffect`. This is to allow the modal to retrigger during navigation to `/widget/subscribe`

bkrem: My bad, didn't realise it was a dep array at first 🤦‍♂️

bkrem (approved):

PR #183, opened 2023-07-27
Bubble up the push message event

devceline:
# Description

- Bubble push message event message to window parent if url matches dappOrigin

# Type of change

# Checklist:

### DIFF
--- a/src/contexts/W3iContext/hooks/pushHooks.ts
+++ b/src/contexts/W3iContext/hooks/pushHooks.ts
@@ -5,8 +5,11 @@ import { noop } from 'rxjs'
 import { subscribeModalService } from '../../../utils/store'
 import type { W3iPushClient } from '../../../w3iProxy'
 import type Web3InboxProxy from '../../../w3iProxy'
+import { JsCommunicator } from '../../../w3iProxy/externalCommunicators/jsCommunicator'
 import { useAuthState } from './authHooks'
+import { useDappOrigin } from './dappOrigin'
 import { useUiState } from './uiHooks'
+import { EventEmitter } from 'events'
 
 export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean) => {
   const [activeSubscriptions, setActiveSubscriptions] = useState<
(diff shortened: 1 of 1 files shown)


src/contexts/W3iContext/hooks/pushHooks.ts
```diff
@@ -112,5 +118,32 @@ export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean) => {
...
   }, [pushClient, refreshPushState])
 
+  useEffect(() => {
+    if (!pushClient) {
+      return noop
+    }
+
+    const pushMessageSub = pushClient.observe('push_message', {
+      next: message => {
+        if (message.params.message.url !== dappOrigin) {
```
bkrem, line 128: Curious: do we need this because this observer is also in place in non-dappOrigin scenarios? Or why the safeguard if the client should only be subscribed to push topics relevant to this dapp?

src/contexts/W3iContext/hooks/pushHooks.ts
```diff
@@ -112,5 +118,32 @@ export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean) => {
...
+          return
+        }
+
+        const communicator = new JsCommunicator(emitter)
+        communicator.postToExternalProvider(
+          'dappPushNotification',
```
bkrem, line 134: Is this naming based on something that's been agreed? We seem to be mixing and matching camelCase and snake case on these calls too: https://github.com/WalletConnect/web3inbox/blob/main/src/w3iProxy/index.ts#L87

src/contexts/W3iContext/hooks/pushHooks.ts
bkrem: Would be great to add a comment explaining intent on top either way 🙏 

There's a lot going on between w3i standalone + the dapp widget integration so we should aim to make the context clearer over time for others regarding why specific checks and/or sub-routines are in place.

bkrem (approved):

src/contexts/W3iContext/hooks/pushHooks.ts
devceline: it should be underscored, I'll do that

src/contexts/W3iContext/hooks/pushHooks.ts
bkrem: Great description of the why, ty 🙏

PR #195, opened 2023-08-10
Change policy injection

devceline:
# Description

- Change register to not inject index file, this achieves two things:
    - First, we lose offline work. Not a big loss since our app really needs a connection.
    - Second, we no longer get an issue where we get an outdated index file that fetches an incorrect bundle name
    - This just makes the service worker a way to more efficiently catch pictures and the like


Cali93 (approved):
LGTM 👍

PR #197, opened 2023-08-17
chore(notify-register): replace enableSync by register

Cali93:


src/contexts/W3iContext/hooks/pushHooks.ts
```diff
@@ -47,12 +47,13 @@ export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean, dapp
 
   const handleRegistration = useCallback(
     async (key: string) => {
+      console.log({ pushClient, key, pushEnabled: uiEnabled.notify })
```
devceline, line 50: Do we need this log?

src/contexts/W3iContext/hooks/pushHooks.ts
```diff
@@ -47,12 +47,13 @@ export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean, dapp
...
       if (pushClient && key && uiEnabled.notify) {
         try {
           const identityKey = await pushClient.register({ account: `eip155:1:${key}` })
+          refreshPushState()
           setRegisterMessage(null)
           setRegistered(identityKey)
-          refreshPushState()
```
devceline, line 53: What's the reasoning for this?

devceline (approved):
LGTM

src/contexts/W3iContext/hooks/pushHooks.ts
Cali93: I looked into how `chatHooks` handle it and I used the same order, curious if it matters ?

PR #205, opened 2023-08-24
Fix low-hanging fruit

devceline:
# Description
- Use `JSX` compliant properties
- Sort notifications
- Account for sync init in `pushHooks`
  - Note, current solution for this is a stopgap, we need to look into creating events for this. 
# Type of change

# Checklist:

### DIFF
--- a/src/contexts/W3iContext/hooks/pushHooks.ts
+++ b/src/contexts/W3iContext/hooks/pushHooks.ts
@@ -42,7 +42,10 @@ export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean, dapp
   }, [pushClient, userPubkey])
 
   useEffect(() => {
-    refreshPushState()
+    // Account for sync init
+    const timeoutId = setTimeout(() => refreshPushState(), 100)
+
+    return () => clearTimeout(timeoutId)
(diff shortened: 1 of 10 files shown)


src/contexts/W3iContext/hooks/pushHooks.ts
```diff
@@ -42,7 +42,10 @@ export const usePushState = (w3iProxy: Web3InboxProxy, proxyReady: boolean, dapp
   }, [pushClient, userPubkey])
 
   useEffect(() => {
-    refreshPushState()
+    // Account for sync init
+    const timeoutId = setTimeout(() => refreshPushState(), 100)
```
bkrem, line 46: What event are we missing to handle this properly?

bkrem (approved):

src/contexts/W3iContext/hooks/pushHooks.ts
devceline: We need something like `sync_stores_initialized` that accounts for the init of `subscriptions`. This is due to the fact that way stores are initialized is we first init a `Store` instance for `subscriptions`. Then, after sync keys are updated etc we initialize a `SyncStore` instance. There is no event for that

PR #206, opened 2023-08-24
Revert "fix: resolve `.well-known` assets via app.browser field for testing

bkrem:
This reverts commit 1b6688d18f2700b22aa2cdf1b96a58d14062e238.

# Description

- This was a workaround/override behaviour we used on the notify refactor branch to specifically target `notify.gm.walletconnect.com` in the w3i explorer
- `gm-dapp`'s homepage field has now been updated so this is no longer needed, i.e. we should always point to the `homepage` field going forward as before.


devceline (approved):
LGTM

PR #211, opened 2023-08-24
Fix/init push sync stores

devceline:
# Description
## Sync Store Initializes
- Allow notify sync stores to be initialized once chat's sync stores are initialized. 
- Use event based method instead of `setTimeout`
## Init Data races
- `w3iProxy` was being initted *twice* by the `useEffect` in `w3iProxyHooks`.  Solved this by adding `initializing` state to the proxy
## Chat
- Chat's `register` method was sometimes failing to register invites. This error has been silenced and ignored. Function's API of returning identity key is maintained. However, registering invite key is not working and it shouldn't stop w3i from working at the moment.
- This was not caught before because of the following:
  - W3iProxy would run, attempt to initialize chat client. Succeed. Attempt to register. Fail and die
  - Second instance of W3iProxy (due to the above issue of it being initted twice) would initiate `notify` and `auth`

# Type of change

# Checklist:

### DIFF
--- a/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
+++ b/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
@@ -28,14 +28,13 @@ export const useW3iProxy = () => {
   )
 
   useEffect(() => {
-    setReady(oldReady => {
-      if (!oldReady) {
-        w3iProxy.init()
-      }
-
-      return true
-    })
-  }, [w3iProxy, setReady])
+    if (w3iProxy.isInitializing) {
+      return
+    }
+    if (!w3iProxy.getInitComplete()) {
+      w3iProxy.init().then(() => setReady(true))
+    }
+  }, [w3iProxy.isInitializing])
 
   return [w3iProxy, ready] as [Web3InboxProxy, boolean]
 }
--- a/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
+++ b/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
@@ -28,7 +28,13 @@ export const useW3iProxy = () => {
   )
 
   useEffect(() => {
-    w3iProxy.init().then(() => setReady(true))
+    setReady(oldReady => {
+      if (!oldReady) {
+        w3iProxy.init()
+      }
+
+      return true
+    })
   }, [w3iProxy, setReady])
 
   return [w3iProxy, ready] as [Web3InboxProxy, boolean]
--- a/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
+++ b/src/contexts/W3iContext/hooks/w3iProxyHooks.ts
@@ -32,7 +32,7 @@ export const useW3iProxy = () => {
       return
     }
     if (!w3iProxy.getInitComplete()) {
-      w3iProxy.init().then(() => setReady(true))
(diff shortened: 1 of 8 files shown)


src/w3iProxy/index.ts
```diff
@@ -181,23 +187,41 @@ class Web3InboxProxy {
...
+      if (this.chatClient) {
+        this.chatClient.once('sync_stores_initialized', () => {
+          const account = this.auth.getAccount()
+          if (this.pushClient && account) {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const signature = this.chatClient!.syncClient!.signatures.get(
+              `eip155:1:${account}`
+            )!.signature
+
+            this.pushClient.initSyncStores({
```
bkrem, line 214: > Allow notify sync stores to be initialized once chat's sync stores are initialized.

Why do we need do make this coupled between the two independent clients here? Basically I'm not following how this relates back to the original issue you described around duplicate spawning of `w3iProxy` and/or silent failure of chat registration.

src/w3iProxy/chatProviders/internalChatProvider.ts
```diff
@@ -242,21 +242,30 @@ export default class InternalChatProvider implements W3iChatProvider {
...
-        })
-      }
-    })
+        }
+      })
+
+      return registeredIdentityKey
+    } catch (e) {
+      /*
+       * Right now, chat client's register doesn't work 100% of the time
```
bkrem, line 262: Thanks for the explanation and the todo. Let's please create an issue in chatClient 🙏 

Going to be a lot more watchful of our error handling in w3i in reviews going forward, unhandled promise rejections have caused hard-to-understand issues a bunch of times now, primarily wasting your time in particular in the process.

src/contexts/W3iContext/hooks/w3iProxyHooks.ts
```diff
@@ -28,8 +28,13 @@ export const useW3iProxy = () => {
   )
 
   useEffect(() => {
-    w3iProxy.init().then(() => setReady(true))
-  }, [w3iProxy, setReady])
+    if (w3iProxy.isInitializing) {
+      return
+    }
+    if (!w3iProxy.getInitComplete()) {
+      w3iProxy.init().then(() => setReady(true))
```
bkrem, line 35: While on the note of explicit error handling:

```suggestion
      w3iProxy.init().then(() => setReady(true)).catch(error => { console.error("w3iProxy failed to initialize: ", error) })
```

bkrem (approved):
Great job identifying the underlying issue and really nice description of what it was and how it's being resolved 💪 🙏

src/w3iProxy/index.ts
devceline: This was the original purpose of this PR before it became a rabbit hole of multiple bug fixes. When you start with a clean slate (`localStorage.clear()`) `this.pushClient.register` never fires because `chatEnabled=true`. 

This mitigates that by not calling `.register` and just initting the sync stores, which is what we care about.

src/w3iProxy/chatProviders/internalChatProvider.ts
devceline: Likewise, if it was properly logged that w3i init was throwing this would've been caught ages ago. Thanks for the below suggestion

devceline: https://github.com/WalletConnect/chat-client-js/issues/68

PR #214, opened 2023-08-28
Add failsafe

devceline:
# Description

- Adding a couple of failsafes to account for the rare occurrence where the widget subscribe page does not detect if there's an active sub

# Type of change

# Checklist:

### DIFF
--- a/src/pages/widget/Subscribe/index.tsx
+++ b/src/pages/widget/Subscribe/index.tsx
@@ -21,6 +21,7 @@ const WidgetSubscribe: React.FC = () => {
   } = useContext(W3iContext)
 
   const nav = useNavigate()
+  const [emitter] = useState(new EventEmitter())
 
   const [isSubscribing, setIsSubscribing] = useState(false)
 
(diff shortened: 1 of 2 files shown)


src/pages/widget/Subscribe/index.tsx
```diff
@@ -52,17 +53,29 @@ const WidgetSubscribe: React.FC = () => {
...
         nav(`/notifications/${dappSub.topic}`)
-      }, 0)
+      }
     }
-  }, [activeSubscriptions, nav])
+
+    navIfSubFound()
+
+    /*
+     * The following two event listeners are essentially fail-safes if something goes wrong
```
bkrem, line 68: What is "something goes wrong" in this case and do we know why? I'd prefer to figure out root cause since sticking on a band aid isn't the way forward long term.

src/w3iProxy/pushProviders/internalPushProvider.ts
```diff
@@ -118,6 +118,24 @@ export default class InternalPushProvider implements W3iPushProvider {
       throw new Error(this.formatClientRelatedError('subscribe'))
     }
     console.log('InternalPushProvider > PushClient.subscribe > params', params)
+    const currentSubs = Object.values(
+      this.pushClient.getActiveSubscriptions({ account: params.account })
+    )
+
+    /*
+     * We do nothing with this data so it's okay if we return garbage, for now.
```
bkrem, line 126: Why do we need this if the data isn't used? This seems to also be creating a mixed return type here which is either `boolean` or this object on the early return?

Again feels like a hack. I see "Move this functionality into notify client." below, but what do we have to add/what does the issue for this look like?

devceline: It always returns 
```ts
{
  id: string,
  subscribeAuth: string
}
```

I made it return the empty object. I've kept this API to remain compliant with notify client's API. 

I'm still unsure how we should handle it in notify client. Like the comment says, we can't be confident we have the request id, and we definitely can't recreate the original subscription auth. 

Maybe we can make it throw if you try to subscribe to existing sub and handle this in a try catch? We'd still need to return something though.

The other option is having w3i not comply with notify client's API, and return a boolean. I'm open to ideas

devceline:
Closing this in favor of a more sustainable solution

PR #216, opened 2023-08-30
Fix not disconnecting

devceline:
# Description

- Fix issue where sometimes disconnecting does not work
- It was caused by navigating away or changing state that would trigger navigating away before disconnecting finalizes. 
- Both these instances were fixed.
- Add temporary band-aid until wagmi updates their ethereum provider
# Type of change

# Checklist:

This is still in draft because while fixing this I stumbled upon a different issue where sometimes the session does not get cleaned up, leaving behind a broken session. This prevents connecting again


bkrem (approved):
Sorry started reviewing this earlier today and clearly got distracted with IRL workshops and Cloud topics 🙏

PR #218, opened 2023-08-31
Add polling to account for history injection

devceline:
# Description

There is no real way to know when history is done being injected since it emits different events based on fetched messages. Maybe down the line we could add a mechanism so that after fetching from history, we decode messages and add a counter.

However, this is going to take a lot of effort for something than can be achieved with polling

# Type of change

# Checklist:


bkrem:
@devceline Can we close?

PR #220, opened 2023-09-01
Add extensive logging through mixpanel

devceline:
# Description
- Use mixpanel with logging by transmitting
# Type of change

# Checklist:

This forgoes the need for a `transport` module.


package.json
```diff
@@ -29,6 +29,7 @@
     "framer-motion": "^10.3.0",
     "idb": "^7.1.1",
     "lodash.debounce": "^4.0.8",
+    "mixpanel-browser": "^2.47.0",
```
bkrem, line 32: So the custom transport wasn't needed in the end?

src/w3iProxy/index.ts
```diff
@@ -163,6 +183,13 @@ class Web3InboxProxy {
 
     this.initializing = true
 
+    if (this.core) {
```
bkrem, line 186: Can we move this to here (https://github.com/WalletConnect/web3inbox/pull/220/files#diff-82bbad320eb11fb8fe700ee91086488459dc806d83d3298899fbecf6ab15da64R96) where the core is created conditionally, instead of creating a separate if check?

bkrem (approved):
💪 💪

package.json
devceline: Yes, looked closer at the docs and found a simpler/cleaner way

src/w3iProxy/index.ts
devceline: It would end being a long, kind of unreadable promise chain so I went with this. If this is a strong preference I can do that

PR #238, opened 2023-09-26
Remove chat client

devceline:
# Description

- Remove chat client
- This will save us maintenance effort in the long run, as we will no longer have to bump chat client with every core update etc. 
- Also, this completely disables chat UI and one would have to go out of their way, modifying source code just to see it. So now it is clearly deprecated/not for public use. 
- Finally, with the removal of chat client from init logic, we can reduce surface area for bugs and just focus on creating a great notify experience
- Most of this code would _probably_ be changed anyway so there is not much point keeping it around commented. 
- If need be, we can go back to this PR and look at the commits. Having commented code would just make it a hassle to navigate the actual code.
# Type of change

- [x] Chore

# Checklist:


package.json
```diff
@@ -48,6 +46,7 @@
     "@typescript-eslint/eslint-plugin": "^5.45.1",
     "@typescript-eslint/parser": "^5.45.1",
     "@vitejs/plugin-react": "^2.2.0",
+    "@walletconnect/chat-client": "^0.8.2",
```
bkrem, line 49: Accidental add as devDep?

src/components/messages/ThreadSelector/Thread/index.tsx
```diff
@@ -19,20 +18,21 @@ const Thread: React.FC<ThreadProps> = ({
...
-      chatClientProxy?.getMessages({ topic }).then(messages => {
-        if (messages.length) {
-          setCalculatedLastMessage(messages[messages.length - 1].message)
-          setCalculatedLastMsgTimestamp(messages[messages.length - 1].timestamp)
-        }
-      })
-    }
-  }, [chatClientProxy, calculatedLastMessage, lastMessage, setCalculatedLastMessage])
+  /*
+   * UseEffect(() => {
```
bkrem, line 25: Can we remove too rather than comment out?

src/components/notifications/NotificationsLayout/index.tsx
```diff
@@ -5,6 +5,8 @@ import { AnimatePresence } from 'framer-motion'
 import { motion } from 'framer-motion'
 
 const NotificationsLayout: React.FC = () => {
+  console.log('>>>> aa')
```
bkrem, line 8:
```suggestion

```

src/contexts/W3iContext/index.tsx
```diff
@@ -41,28 +31,25 @@ const W3iContextProvider: React.FC<W3iContextProviderProps> = ({ children }) =>
...
     <W3iContext.Provider
       value={{
-        chatClientProxy: chatClient,
-        chatProvider,
+        chatClientProxy: null,
         pushProvider,
         authProvider,
         userPubkey,
-        uiEnabled,
+        uiEnabled: { chat: false, ...uiEnabled },
```
bkrem, line 38: If `uiEnabled` still contains `chat: true` for some reason it will override the explicit setting based on the way the rest operator behaves if it comes after the explicit setting:

```suggestion
        uiEnabled: { ...uiEnabled, chat: false },
```

src/pages/Login/SignatureModal/index.tsx
```diff
@@ -25,9 +25,7 @@ export const SignatureModal: React.FC<{
       .then(signature => {
         switch (sender) {
           case 'chat':
-            window.web3inbox.chat.postMessage(
-              formatJsonRpcRequest('chat_signature_delivered', { signature })
-            )
+	    console.warn("signing messages for chat is not supported")
```
bkrem, line 28:
```suggestion
            console.warn("[Web3Inbox] Signing messages for chat is not supported.")
```

bkrem: Looks like a tab got mixed in here too. Tried to fix in the suggestion.

bkrem (approved):
Some minor comments but LGTM ✨ 

Played around on the preview and looks fine to me there. Please double check on your end too if you haven't yet since we don't have a dedicated staging env/branch yet 🙏

bkrem:
Also nice context on the PR description 💯

package.json
devceline: No - It's still used to build the types. Considering hard coding the type but not sure if it's the right approach

devceline: Decided it wasn't worth the hassle and it's cleaner to hard code the types. But this required some additions.

src/pages/Login/SignatureModal/index.tsx
```diff
@@ -25,7 +25,7 @@ export const SignatureModal: React.FC<{
       .then(signature => {
         switch (sender) {
           case 'chat':
-	    console.warn("signing messages for chat is not supported")
+        console.warn("[Web3Inbox] Signing messages for chat is not supported.")
```
bkrem, line 28: Damn, didn't get indentation right the first time 😄 

```suggestion
            console.warn("[Web3Inbox] Signing messages for chat is not supported.")
```

bkrem (approved):

PR #250, opened 2023-10-10
feat: refactor auth address prefix

### LINKED ISSUE #248: Don't use hardcoded eip155:1, be chain aware
Web3inbox currently assumes `eip155:1` , it needs to be chain aware

Web3inbox's current `authProvider` does not keep track of chains, the rest of the codebase simply assumes `eip155:1` is always the chain.

To achieve this:
- refactor `w3iAuthFacade`, `internalAuthProvider` and `externalAuthProvider` to keep track of chain/networek
- align the rest of the codebase to not assume `eip155:1` and instead take the chain from the auth provider.
- fix areas where an address is assumed to not have a prefix, eg when `userPubkey` is used directly to fetch data from ens

Results from this:
- Correct subscriptions for account are retrieved, eg logging in with Binance smart chain does not fetch data for mainnet. 
- Subscribe and register from correct address, not always mainnet

devceline:
# Description
- Add chain data to `w3iAuthFacade`
- React to chain switching in `internalAuthProvider` 
- Allow passing chain data to `externalAuthProvider`
- Update web3modal & its config
- **Preliminary list of chains, not final**
- Adapt signature modal trigger to work with chain switching
- Remove all hard coding of `eip155:1` (except for chat components, those will be fixed at a later date)
- Remove unreferenced/dead code that communicated with parent dapp (old `iframe` widget)
# Type of change

# Fixes/Resolves (Optional)

Resolves #248 

# Checklist:

# Testing 

- Connect via WalletConnect using a mobile wallet
- Subscribe on ethereum chain to gm dapp
- Switch to polygon chain
- ensure there is no sub to gm dapp
- subscribe to gm dapp
- switch back to ethereum, ensure original gm dapp sub is there (via topic)
- switch to a different chain try again

### DIFF
--- a/src/components/layout/MobileHeader/index.tsx
+++ b/src/components/layout/MobileHeader/index.tsx
@@ -7,6 +7,7 @@ import Text from '../../general/Text'
 import W3iContext from '../../../contexts/W3iContext/context'
 import ArrowLeftIcon from '../../general/Icon/ArrowLeftIcon'
 import AppNotificationDropdown from '../../notifications/AppNotifications/AppNotificationDropdown'
+import { getEthChainAddress } from '../../../utils/address'
 
 interface IMobileHeaderProps {
   title: string
@@ -48,7 +49,7 @@ const MobileHeader: React.FC<IMobileHeaderProps> = ({ title, back, notificationI
           notificationId={notificationId}
         />
       ) : (
-        <Avatar address={userPubkey as `0x${string}`} width="1.875em" height="1.875em" />
+        <Avatar address={getEthChainAddress(userPubkey) as `0x${string}`} width="1.875em" height="1.875em" />
       )}
     </div>
   )
--- a/src/components/layout/MobileHeader/index.tsx
+++ b/src/components/layout/MobileHeader/index.tsx
@@ -49,7 +49,11 @@ const MobileHeader: React.FC<IMobileHeaderProps> = ({ title, back, notificationI
           notificationId={notificationId}
         />
       ) : (
-        <Avatar address={getEthChainAddress(userPubkey) as `0x${string}`} width="1.875em" height="1.875em" />
+        <Avatar
+          address={getEthChainAddress(userPubkey) as `0x${string}`}
+          width="1.875em"
+          height="1.875em"
+        />
       )}
     </div>
   )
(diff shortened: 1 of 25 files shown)


src/components/notifications/AppExplorer/AppCard/index.tsx
```diff
@@ -35,7 +35,11 @@ const AppCard: React.FC<AppCardProps> = ({ name, description, logo, bgColor, url
 
   const { activeSubscriptions } = useContext(W3iContext)
 
-  const subscribed = activeSubscriptions.some(element => element.metadata.name === name)
+  useEffect(() => {
+    setSubscribing(false);
+  }, [userPubkey])
```
Cali93, line 40: Can we be more explicit about what's going on here ?
```suggestion
  useEffect(() => {
    if(userPubkey) setSubscribing(false);
  }, [userPubkey])
```

src/components/layout/MobileHeader/index.tsx
```diff
@@ -48,7 +49,7 @@ const MobileHeader: React.FC<IMobileHeaderProps> = ({ title, back, notificationI
           notificationId={notificationId}
         />
       ) : (
-        <Avatar address={userPubkey as `0x${string}`} width="1.875em" height="1.875em" />
+        <Avatar address={getEthChainAddress(userPubkey) as `0x${string}`} width="1.875em" height="1.875em" />
```
Cali93, line 52: This bit is non-DRY. Should we add it in the context or something ?

Cali93 (approved):
LGTM

src/contexts/W3iContext/hooks/authHooks.ts
```diff
@@ -26,21 +28,25 @@ export const useAuthState = (w3iProxy: Web3InboxProxy, proxyReady: boolean) => {
 
   useEffect(() => {
     if (accountQueryParam && authClient) {
-      authClient.setAccount(accountQueryParam)
+      const splitAccount = accountQueryParam.split(':');
+      if(splitAccount.length === 3) {
+	authClient.updateFullAccount(splitAccount.slice(0, 2).join(':'), splitAccount[2])
+      }
```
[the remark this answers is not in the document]
devceline: good idea, yep will do

src/components/notifications/AppExplorer/AppCard/index.tsx
devceline: No - `setSubscribing(false` should run even if `userPubKey` is falsy. I'll add a clarifying comment though

src/components/layout/MobileHeader/index.tsx
devceline: I'd rather not have two separate states that basicallty just hold address information

PR #276, opened 2023-11-17
chore: return show notification function

devceline:
# Description
**TL;DR: There's a race condition that causes the push notifications to be considered "silent push messages" which are flagged by the browser/OS as suspicious. Causing odd messages before every PN on desktop ("This site has been upgraded in the background") and lack of reliability on iOS. This should fix it.**

According to the sources below, not awaiting the `showNotification` and technically firing it after the listener resolves is a red flag for the browser/OS. 

- [StackOverflow, Chrome PNs causing odd messages "Site has been upgraded in the background"](https://stackoverflow.com/questions/31108699/chrome-push-notification-this-site-has-been-updated-in-the-background)
- [web.dev, handling push notifications](https://web.dev/articles/push-notifications-handling-messages#:~:text=Calling%20self.,the%20notification%20has%20been%20displayed.)
- [TruePush, iOS does not support silent web push messages](https://www.truepush.com/blog/ios-pwa-push-notifications/)
^ The means that a notification has to be shown to every `push` event. If the registration's `showMessage` is fired after that is resolved, it is considered a silent message. Red flag for iOS

**This also fixes issue causing white screens, by guarding every single reference to `Notification` even when it logically wouldn't be accessed.  Due to "Variable Not Found Errors".**

[Slack thread about Notification variable causing white screens](https://walletconnect.slack.com/archives/C0665P2HYM9/p1700230743240109?thread_ts=1700230313.710669&cid=C0665P2HYM9)
# Type of change

# Checklist:

### DIFF
--- a/src/Modals.tsx
+++ b/src/Modals.tsx
@@ -26,7 +26,7 @@ export const Modals = () => {
 
   const notificationsEnabled = useNotificationPermissionState()
 
-  const explicitlyDeniedOnDesktop = !isMobile() && Notification?.permission === 'denied'
+  const explicitlyDeniedOnDesktop = !isMobile() && window.Notification?.permission === 'denied'
 
   const shouldShowNotificationModal = useMemo(
     () =>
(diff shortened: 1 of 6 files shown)


bkrem (approved):

bkrem:
Thanks for references/related reading here to help understand the problem 🙏

PR #277, opened 2023-11-20
fix: redirect when there are no subs

devceline:
# Description
- Run prettier on unformatted files
- Fix a bug in `NotificationsLayout` where user was being redirected to `/new-app` even when they had existing subscriptions (on mobile), which is not the intended experience
- This fixes it to intended UX:
    - On desktop, always go to `/notifications/new-app` when on `/notifications` (as we have the screen space to display the selector an the explorer
    - On mobile, only go `/notifications/new-app` if there are no subscriptions

# Type of change

# Checklist:

### DIFF
--- a/src/components/notifications/NotificationsLayout/index.tsx
+++ b/src/components/notifications/NotificationsLayout/index.tsx
@@ -4,13 +4,15 @@ import AppSelector from '../AppSelector'
 import { AnimatePresence } from 'framer-motion'
 import { motion } from 'framer-motion'
 import W3iContext from '../../../contexts/W3iContext/context'
+import { useIsMobile } from '../../../utils/hooks'
 
 const NotificationsLayout: React.FC = () => {
   const { activeSubscriptions } = useContext(W3iContext)
   const { pathname } = useLocation()
+  const isMobile = useIsMobile()
 
   if (pathname === '/notifications') {
-    if (!activeSubscriptions.length) {
+    if (isMobile && !activeSubscriptions.length) {
       return <Navigate to="/notifications/new-app" />
     }
   }
--- a/src/components/notifications/NotificationsLayout/index.tsx
+++ b/src/components/notifications/NotificationsLayout/index.tsx
@@ -4,15 +4,13 @@ import AppSelector from '../AppSelector'
 import { AnimatePresence } from 'framer-motion'
 import { motion } from 'framer-motion'
 import W3iContext from '../../../contexts/W3iContext/context'
-import { useIsMobile } from '../../../utils/hooks'
 
 const NotificationsLayout: React.FC = () => {
   const { activeSubscriptions } = useContext(W3iContext)
   const { pathname } = useLocation()
-  const isMobile = useIsMobile()
 
   if (pathname === '/notifications') {
-    if (isMobile && !activeSubscriptions.length) {
+    if (!activeSubscriptions.length) {
       return <Navigate to="/notifications/new-app" />
     }
   }
--- a/src/components/notifications/NotificationsLayout/index.tsx
+++ b/src/components/notifications/NotificationsLayout/index.tsx
@@ -1,14 +1,18 @@
-import React, { Fragment } from 'react'
+import React, { Fragment, useContext } from 'react'
 import { Navigate, Outlet, useLocation } from 'react-router-dom'
 import AppSelector from '../AppSelector'
 import { AnimatePresence } from 'framer-motion'
 import { motion } from 'framer-motion'
+import W3iContext from '../../../contexts/W3iContext/context'
 
 const NotificationsLayout: React.FC = () => {
+  const { activeSubscriptions } = useContext(W3iContext)
   const { pathname } = useLocation()
 
   if (pathname === '/notifications') {
-    return <Navigate to="/notifications/new-app" />
+    if (!activeSubscriptions.length) {
+      return <Navigate to="/notifications/new-app" />
+    }
   }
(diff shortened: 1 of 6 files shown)


src/utils/hooks/useNotifyProjects.ts
```diff
@@ -11,15 +11,17 @@ const useNotifyProjects = () => {
       const explorerApiBaseUrl: string = import.meta.env.VITE_EXPLORER_API_URL
       const projectId: string = import.meta.env.VITE_PROJECT_ID
 
-      const explorerUrl = filterAppDomain? `${explorerApiBaseUrl}/w3i/v1/notify-config?projectId=${projectId}&appDomain=${filterAppDomain}` : `${explorerApiBaseUrl}/w3i/v1/projects?projectId=${projectId}&is_verified=${
-        isDevModeEnabled ? 'false' : 'true'
-      }`
+      const explorerUrl = filterAppDomain
+        ? `${explorerApiBaseUrl}/w3i/v1/notify-config?projectId=${projectId}&appDomain=${filterAppDomain}`
```
enesozturk, line 15: For better readability, can we use the `URL` class to create URLs? That would help when we add more params or refactor it.


```js
const explorerUrl = new URL(ExplorerEndpoints.projects, explorerApiBaseUrl)
explorerUrl.searchParams.set('projectId', projectId)
explorerUrl.searchParams.set('is_verified', isDevModeEnabled ? 'false' : 'true')
```

Now we have two different options depending on the filterAppDomain so if we could move these URL generation to a utility function

src/components/notifications/NotificationsLayout/index.tsx
```diff
@@ -1,14 +1,18 @@
...
 import { motion } from 'framer-motion'
+import W3iContext from '../../../contexts/W3iContext/context'
 
 const NotificationsLayout: React.FC = () => {
+  const { activeSubscriptions } = useContext(W3iContext)
   const { pathname } = useLocation()
 
   if (pathname === '/notifications') {
-    return <Navigate to="/notifications/new-app" />
+    if (!activeSubscriptions.length) {
```
enesozturk, line 13: Why do we do this? One thing about the subscriptions we should be consider: Is it's returning an empty array in the first render and this might lead us. 

So eventually in the first time it'll be empty even though I have subscriptions, should we handle this differently?

src/utils/hooks/useNotifyProjects.ts
enesozturk: Here: https://github.com/WalletConnect/web3inbox/pull/278/files#diff-0d91658008e7f6731408fc6871b1da50e422f43d2e582fe942167fe13a89e732R17-R19

I think I could update the logic here once you merge this one

devceline: Good point! Yes it's time we refactor some of this

src/components/notifications/NotificationsLayout/index.tsx
devceline: That's okay for now - maybe down the line we could do something to do with the spinners. 

Maybe something like

- Go to app selector
- Have spinners show up for the subs
- If no subs exist, redirect to discover page
- if they do, show them

src/utils/hooks/useNotifyProjects.ts
```diff
@@ -1,21 +1,26 @@
...
     const fetchNotifyProjects = async () => {
-      const explorerApiBaseUrl: string = import.meta.env.VITE_EXPLORER_API_URL
       const projectId: string = import.meta.env.VITE_PROJECT_ID
 
-      const explorerUrl = filterAppDomain
-        ? `${explorerApiBaseUrl}/w3i/v1/notify-config?projectId=${projectId}&appDomain=${filterAppDomain}`
-        : `${explorerApiBaseUrl}/w3i/v1/projects?projectId=${projectId}&is_verified=${
-            isDevModeEnabled ? 'false' : 'true'
-          }`
+      const explorerUrl = new URL(filterAppDomain? EXPLORER_ENDPOINTS.notifyConfig : EXPLORER_ENDPOINTS.projects, EXPLORER_API_BASE_URL);
```
enesozturk, line 14: Should be file formatting issue here in `filterAppDomain?`

src/components/notifications/NotificationsLayout/index.tsx
enesozturk: @devceline Also we are doing the same redirection here: https://github.com/WalletConnect/web3inbox/blob/main/src/components/notifications/AppSelector/index.tsx#L59-L63

Seems like you added mobile condition. Is this redirection expected only on mobile? Should we remove the `useEffect` here?

devceline: Excellent catch, fixed up the logic. Removed the redirect in `AppSelector`

src/utils/constants.ts
```diff
@@ -3,3 +3,10 @@ export const SERVICE_WORKER_ACTIONS = {
   SET_SUBS_SYMKEYS: 'SET_SUBS_SYMKEYS',
   REGISTER_WITH_ECHO: 'REGISTER_WITH_ECHO'
 }
+
+export const EXPLORER_API_BASE_URL: string = import.meta.env.VITE_EXPLORER_API_URL
+
+export const EXPLORER_ENDPOINTS = {
```
enesozturk, line 9: 💯

enesozturk (approved):
LGTM 🚀

Issue #256, opened 2023-10-11
Improve Notifications Sidebar


herrhaase:
```[tasklist]
### Tasks
- [x] Adjust sidebar length to 320px (fixed) (P2)
- [x] Truncate long dapp names (P1)
- [x] Make hover animation more snappier (P2)
```

herrhaase:
<img width="883" alt="image" src="https://github.com/WalletConnect/web3inbox/assets/707251/2b95312c-28bf-4e2a-bb2f-141aef009511">

Instead of being responsive, the Notifications sidebar should be always **fixed at 320px.** Longer app names should be truncated. 

<img width="381" alt="image" src="https://github.com/WalletConnect/web3inbox/assets/707251/2917338b-7617-49b4-ab8e-243211e38407">

herrhaase:
**Make hover animation more snappier** 

https://github.com/WalletConnect/web3inbox/assets/707251/26ec620c-c1cf-4b02-a79c-089223275b09

Current animation feels quite slow. Can we reduce duration to 300ms and see how that feels?

enesozturk:
@chris13524 @chadyj The tasks in this ticket are also fixed.

chris13524:
LGTM. Closing

PR #295, opened 2023-11-28
refactor: update app selector and hover stylings

enesozturk:
# Description

Truncate dApp list item name as one line and refactored related styles

- https://github.com/WalletConnect/web3inbox/issues/256

# Type of change

# Checklist:


src/App.scss
```diff
@@ -63,28 +64,29 @@
       padding: 1rem 0.085em;
     }
   }
+
   .App {
     --targetselector-display: unset;
     --sidebar-display: unset;
     --header-display: unset;
     --main-display: unset;
-    grid-template: 'sidebar target-selector main';
```
enesozturk, line 71: Grid is a nice approach to creating complex layouts with ease. In our situation, I found this approach not quite modular. Every item (like SideBar, AppSelector, Main, etc) should have its sizing properties on its own implementation. This will make them more movable and maintainable

Since the SideBar and AppSelector have static widths, lets handle their styles on their class and just use flex to position them.

src/components/general/NavLink/NavLink.scss
```diff
@@ -19,7 +19,7 @@
   &:hover,
   &-active {
     @media (hover: hover) {
-      background: rgba(51, 150, 255, 0.08);
+      background: var(--accent-hover);
```
enesozturk, line 22: The variables might needed to be updated in the future but for now, I've set custom color for the dark theme and started using this variable on every hover background.

src/components/notifications/AppSelector/AppSelector.scss
```diff
@@ -161,12 +166,12 @@
       padding: 1rem 1rem 1rem 1rem;
       border-radius: 1em;
       position: relative;
-      transition: all 250ms ease-in-out;
+      transition: background-color 0.1s linear;
```
enesozturk, line 169: While we set the transition property, we should always set which properties will be animated, instead using `all`. 

Also, it's always good for performance practice to use `will-change`. It's a property that let the browser know which props will be animated. And the browser allocates itself depending on this and will help to make the animations more performant.

- https://web.dev/articles/animations-guide

src/components/layout/Sidebar/Sidebar.scss
```diff
@@ -6,7 +6,8 @@
   left: 50%;
   transform: translateX(-50%);
   gap: 1em;
-  height: 3em;
+  width: 60px;
```
devceline, line 9: I'm personally a big fan of relative units, why go with `px` here

src/components/notifications/AppSelector/AppSelector.scss
```diff
@@ -131,6 +131,11 @@
     }
 
     &__title {
+      display: -webkit-box;
+      -webkit-line-clamp: 1;
+      -webkit-box-orient: vertical;
```
devceline, line 136: What's compatibility like for this? Also we have  `autoprefixer`, can we take advantage of that here?

src/components/notifications/AppSelector/AppSelector.scss
devceline: Awesome, always love improving performance

devceline (approved):
LGTM

src/components/layout/Sidebar/Sidebar.scss
enesozturk: Definitely, lets keep using `rem` here.

src/components/notifications/AppSelector/AppSelector.scss
enesozturk: As I see the `autoprefixer` didn't add support for `line-clamp` and `box-orient` yet. It's mentioned in a few issues but I don't see a new release after `10.4.14` which includes those changes. Let's keep using the prefixes here.

Issue #257, opened 2023-10-11
Improve Notifications page


herrhaase:
```[tasklist]
### Tasks
- [ ] Fix font styles (see below) (P2)
- [x] Fix "Show more" (see below) (P0)
- [ ] Fix hover states (see below) (P2)
- [x] Update placeholder icon (see below) (P1)
- [x] Ensure every app has a name (see below)
- [x] Ensure selection style is always on light theme (see below) (P2)
- [ ] Improve empty state (see below) [Update picture to standard, use new copy] (p2)
- [ ] Improve Unsubscribe Modal (p2)
- [ ] Improve settings dropdown (p2)
```

herrhaase:
**Fix font styles**

<img width="850" alt="image" src="https://github.com/WalletConnect/web3inbox/assets/707251/7120c996-7b2e-4ab8-b343-5c5c72c726e6">

See details [here](https://www.figma.com/file/P70QpTDpsgwdPJ2jJmVXxg/Web3Inbox?type=design&node-id=3380%3A3198&mode=design&t=EAfRQa8hP0MiqKeC-1)

herrhaase:
**Fix "Show more"**

1. Doesnt always show up – should be always visible.
2. Doesnt work when clicking on it.

herrhaase:
**Fix hover states**

<img width="1399" alt="image" src="https://github.com/WalletConnect/web3inbox/assets/707251/eda7cc4d-49fa-44e2-8806-34daa9a48468">

herrhaase:
**Update placeholder icon**

<img width="922" alt="image" src="https://github.com/WalletConnect/web3inbox/assets/707251/6b1ec1de-9d35-4d3c-bdd3-8a9150deae1e">

https://www.figma.com/file/P70QpTDpsgwdPJ2jJmVXxg/Web3Inbox?type=design&node-id=3432%3A65755&mode=design&t=ZDISUYI7cDwGsmmq-1

herrhaase:
**Ensure every app has a name**

<img width="1449" alt="image" src="https://github.com/WalletConnect/web3inbox/assets/707251/abf29ed9-3ed9-48a1-a2c0-ce2e07b56668">

Pancake's staging integration currently does not have a name that is displayed. Until Cloud is ready to allow projects to manage their app we should force a name in the UI. This could be as simply as the app's ID.

herrhaase:
**Ensure selection style is always on light theme**

[...]

PR #297, opened 2023-11-29
refactor: update logos with fallback

enesozturk:
# Description

Updated all the logos that have a fallback. I need to remove the `onError` prop since the `onerror` handler is only called it cannot fetch the source. So passing non-string falsy values like `undefined`, and `null` don't trigger the error handler, and that is why it's not updating the src. Instead, src value with optional value of fallback svg path.

# Type of change

# Fixes/Resolves (Optional)

https://github.com/WalletConnect/web3inbox/issues/257#issue-1938393611

# Checklist:


devceline (commented):
There's a visible padding between the border and the logo in the figma, not in this preview though

devceline (approved):
LGTM

PR #304, opened 2023-12-01
refactor: unsubs color and copy

enesozturk:
# Description

Refactored unsubscribe modal and dropdown styles and icons to match the designs

# Type of change

# Checklist:

Fixes: https://github.com/WalletConnect/web3inbox/issues/257#issuecomment-1766375065

### DIFF
--- a/src/components/notifications/AppNotifications/AppNotificationDropdown/AppNotificationDropdown.scss
+++ b/src/components/notifications/AppNotifications/AppNotificationDropdown/AppNotificationDropdown.scss
@@ -38,6 +38,11 @@
       }
     }
 
+    button:focus-visible {
+      outline: none;
+      background-color: var(--bg-color-4);
+    }
+
     button:not(&__unsubscribe) {
       color: var(--fg-color-1);
     }


src/utils/hooks.ts
```diff
@@ -96,7 +96,7 @@ export const useColorModeValue = (mode: SettingsContextSimpleState['mode']) => {
       border2: 'hsla(0, 0%, 0%, 0.1)',
       accent1: 'hsla(211, 100%, 60%, 1)',
       accent2: 'hsla(211, 100%, 60%, 0.08)',
-      error1: 'hsla(5, 85%, 60%, 1)',
+      error1: '#ED4747',
```
devceline, line 99: can we stick to `hsla` format

devceline (approved):
LGTM

src/utils/hooks.ts
enesozturk: Though the same, I think should stick to what the designer uses in the designs. I'm also a fan of HSL, but in this case, when we have different color system on design and code, this will make things hard to maintain. The color system approach should be on designer's responsibility.

PR #309, opened 2023-12-05
fix: discover apps

chris13524:
Change "Discover all apps" to "Discover apps" because we are only showing featured apps, not all apps. [Context](https://walletconnect.slack.com/archives/C044SKFKELR/p1701790341053199).


enesozturk (approved):
LGTM

Issue #290, opened 2023-11-27
P1: Discover page: indicate developer project on Discover page


chris13524:
```[tasklist]
### Tasks
- [ ] Update Discover view: https://walletconnect.slack.com/archives/C044SKFKELR/p1700239547655949?thread_ts=1699976611.338179&cid=C044SKFKELR
```

PR #312, opened 2023-12-08
refactor: developer mode projects UI changes

enesozturk:
# Description

We should show status badges for app cards and list items as `dev mode` or `unapproved`, in both the discover page and the app selector (left side) list.

Currently, due to missing fields returning in the subscribed app from the Notify JS client, we need to match the project's data with the currently subscribed app's data, which is not a good approach. The reason is when we have many subscribed apps, and I only see features apps on the Discover page, I cannot understand if a subscribed app 'which hasn't been featured yet' is verified or not, because I don't have that app in the project list. Only project items have the `isVerified` field.

We should return the verification status in the metadata object: https://github.com/WalletConnect/notify-client-js/blob/45a5ac337ced61ca6cdb2a1d535f205873f23dd4/packages/notify-client/src/controllers/engine.ts#L995

# Type of change

# Checklist:

closes https://github.com/WalletConnect/web3inbox/issues/290


src/components/notifications/AppExplorer/AppCard/index.tsx
```diff
@@ -95,6 +98,9 @@ const AppCard: React.FC<AppCardProps> = ({ name, description, logo, bgColor, url
       <div className="AppCard__header">
         <div className="AppCard__header__logo">
           <img src={logo || '/fallback.svg'} alt={`${name} logo`} />
+          {!isVerified ? (
+            <img src={SpannerSVG} className="AppCard__header__logo__dev-icon" alt="Dev mode icon" />
+          ) : null}
```
Cali93, line 103: Nit pick: you don't need a ternary in this case.
```suggestion
          {!isVerified && (
            <img src={SpannerSVG} className="AppCard__header__logo__dev-icon" alt="Dev mode icon" />
          )}
```

src/components/notifications/AppExplorer/AppCard/index.tsx
```diff
@@ -106,9 +112,8 @@ const AppCard: React.FC<AppCardProps> = ({ name, description, logo, bgColor, url
       </div>
       <div className="AppCard__body">
         <div className="AppCard__body__title">
-          <Text className="" variant="large-600">
-            {name}
-          </Text>
+          <Text variant="large-600">{name}</Text>
+          {!isVerified ? <Badge>DEV</Badge> : null}
```
Cali93, line 116:
```suggestion
          {!isVerified && <Badge>DEV</Badge>}
```

src/components/notifications/AppSelector/index.tsx
```diff
@@ -124,47 +130,59 @@ const AppSelector: React.FC = () => {
           )}
         </div>
         <div className="AppSelector__wrapper">
-          <Label color="main">Subscribed</Label>
+          {!loading && filteredApps?.length > 0 ? <Label color="main">Subscribed</Label> : null}
```
Cali93, line 133: This should work too
```suggestion
          {!loading && filteredApps?.length && <Label color="main">Subscribed</Label>}
```

src/components/notifications/AppSelector/index.tsx
```diff
@@ -124,47 +130,59 @@ const AppSelector: React.FC = () => {
...
-
-                          <div className="AppSelector__link__wrapper">
-                            <Text className="AppSelector__link__title" variant="small-500">
-                              {app.metadata.name}
-                            </Text>
-                            <Text className="AppSelector__link__subtitle" variant="small-500">
-                              {app.metadata.description}
-                            </Text>
+              filteredApps?.map(app => {
+                const unverified = isUnVerified(app, projects)
```
Cali93, line 142: Nit pick: prefix the variable with `is` to denote it's a boolean and rename the function.
```suggestion
                const isUnverified = checkIsUnVerified(app, projects)
```

src/components/notifications/AppSelector/index.tsx
```diff
@@ -124,47 +130,59 @@ const AppSelector: React.FC = () => {
...
+
+                return (
+                  <AnimatePresence>
+                    <m.div initial={{ opacity: 0 }} exit={{ opacity: 0 }} animate={{ opacity: 1 }}>
+                      <NavLink
+                        key={app.topic}
+                        to={NAVIGATION.notifications.topic(app.topic)}
```
Cali93, line 149: It'd be nice to use a more explicit name
```suggestion
                        to={NAVIGATION.notifications.getPathByTopic(app.topic)}
```

src/utils/project.ts
```diff
@@ -0,0 +1,10 @@
+import { NotifyClientTypes } from '@walletconnect/notify-client'
+
+import { INotifyApp } from '@/utils/types'
+
+export function isUnVerified(app: NotifyClientTypes.NotifySubscription, projects: INotifyApp[]) {
+  const project = projects.find(project => project.url.includes(app.metadata.appDomain))
+  const isVerified = project && project?.isVerified !== true
+
+  return isVerified
```
Cali93, line 9: You could make this a one liner using `some`
```suggestion
  return projects.some(project => project.url.includes(app.metadata.appDomain) && project.isVerified !== true)
```

Cali93 (approved):
LGTM, I just have very minor suggestions.
Good job 💪

src/components/notifications/AppSelector/index.tsx
enesozturk: The goal of naming here is to have the same function name as the pathname to match fn names with variable names:

```js
/notifications -> NAVIGATION.notifications
/notifications/{topic} -> NAVIGATION.notifications.topic("TT")
```

When we have inner pages under `/notifications/{topic/*` this might make more sense in terms of readability:

```js
/notifications/{topic}/settings -> NAVIGATION.notifications.topic("TT").settings
/notifications/{topic}/settings/billing -> NAVIGATION.notifications.topic("TT").settings.billing
```

`getPathByTopic ` still make sense tho

enesozturk:
> LGTM, I just have very minor suggestions. Good job 💪

@Cali93 Thanks for the review! 🙏 I've made some updates. For the `&&` and ternary operator reviews, put a comment in the previous PR: https://github.com/WalletConnect/web3inbox/pull/318#discussion_r1441550883

src/components/notifications/AppSelector/index.tsx
Cali93: Thanks for the context, it's a bit unclear what the `topic` function does just by its name, that's why I suggested to rename it.

PR #313, opened 2023-12-17
chore: initial Playwright tests

chris13524:
Adds Playwright testing. Initially created with `npm init playwright@latest` and then tweaks for yarn, automatic running of the dev server, and added a basic initial test.

- [x] Replace with secret before merging

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,9 +1,9 @@
 name: Playwright Tests
 on:
   push:
-    branches: [ main, master ]
+    branches: [ main ]
   pull_request:
-    branches: [ main, master ]
+    branches: [ main ]
 jobs:
   test:
     timeout-minutes: 10
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Install Playwright Browsers
       run: yarn playwright:install
     - name: Run Playwright tests
-      run: yarn test
+      run: yarn playwright:test
       env:
         VITE_PROJECT_ID: ${{ secrets.VITE_DEV_PROJECT_ID }}
     - uses: actions/upload-artifact@v3
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,29 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
+      run: yarn
+    - name: Install Playwright Browsers
+      run: yarn playwright:install
+    - name: Run Playwright tests
+      run: yarn test
+      env:
+        VITE_PROJECT_ID: ${{ secrets.VITE_DEV_PROJECT_ID }}
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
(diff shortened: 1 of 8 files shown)


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,30 @@
...
+        node-version: 18
+    - name: Install dependencies
+      run: yarn
+    - name: Install Playwright Browsers
+      run: yarn playwright:install
+    - name: Run Playwright tests
+      run: yarn test
+      env:
+        # VITE_PROJECT_ID: ${{ secrets.VITE_PROJECT_ID }}
+        VITE_PROJECT_ID: 547aafa48826c4d76f492efecde4843d
```
arein, line 24: what's the game plan here?

arein (approved):

.github/workflows/playwright.yml
chris13524: Will [update the secret](https://walletconnect.slack.com/archives/C03RME3BS9L/p1702856517094549) and then force push to erase this. Pushed temporarily in public since it isn't critically private info, but this really is a secret/variable.

playwright.config.ts
```diff
@@ -0,0 +1,81 @@
+import { defineConfig, devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
```
arein, line 7: consider removing

chris13524: I'd kinda rather keep it for now since this is early in the usage of Playwright and it might come in handy later. Happy for an engineer to remove/cleanup later themselves if they don't feel they will need it.

package.json
```diff
@@ -12,7 +12,9 @@
     "lint": "eslint .",
     "prepare": "husky install",
     "prettier": "prettier --check '**/*.{js,ts,jsx,tsx,scss}'",
-    "prettier:write": "prettier --write '**/*.{js,ts,jsx,tsx,scss}'"
+    "prettier:write": "prettier --write '**/*.{js,ts,jsx,tsx,scss}'",
+    "playwright:install": "playwright install --with-deps",
+    "test": "playwright test"
```
devceline, line 17: We'll also need unit tests at some point as well, so let's make this
```suggestion
    "test:integration": "playwright test"
```
Since it'll be testing w3i like an integration test would, end to end.

playwright.config.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    /* Test against mobile viewports. */
+    // {
+    //   name: 'Mobile Chrome',
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
```
devceline, line 62: Why is this commented out? Mobile is an important use case. 

Do they require specific runners or something?

playwright.config.ts
```diff
@@ -0,0 +1,81 @@
...
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
+    //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
```
devceline, line 73: No need to test for these if we test against chromium

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,29 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
```
devceline, line 4:
```suggestion
    branches: [ main ]
```

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,29 @@
...
+  pull_request:
+    branches: [ main, master ]
```
devceline, line 6:
```suggestion
    branches: [ main ]
```

playwright.config.ts
chris13524: These commented out code blocks are just boilerplate what Playwright init sets up, I haven't enabled them in order to test them. Feel free to enable in a subsequent PR.

playwright.config.ts
chris13524: Same as https://github.com/WalletConnect/web3inbox/pull/313/files#r1429954362

Issue #281, opened 2023-11-22
P2: After unsubscribing from everything, page is blank.


chris13524:

enesozturk:
@chris13524 Can you reproduce this again?

herrhaase:
@chris13524 @enesozturk As soon as a user unsubscribes, we bring them back to the Discover page.

This page is our default starting point whenever there is 
- no active subscription
- no last subscription
- etc.

PR #318, opened 2024-01-03
chore: update subscribed title logic

enesozturk:
# Description

The subscribed title on the app selector panel should be invisible if there is no subscription.

# Type of change

# Fixes/Resolves (Optional)

Closes https://github.com/WalletConnect/web3inbox/issues/281

### DIFF
--- a/src/components/notifications/AppSelector/index.tsx
+++ b/src/components/notifications/AppSelector/index.tsx
@@ -32,6 +32,8 @@ const AppSelector: React.FC = () => {
   const [loading, setLoading] = useState(true)
   const { activeSubscriptions } = useContext(W3iContext)
 
+  const empty = !loading && filteredApps.length === 0
+
   const fetchApps = async (searchQuery: string) => {
     const newFilteredApps = [] as NotifyClientTypes.NotifySubscription[]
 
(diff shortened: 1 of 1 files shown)


src/components/notifications/AppSelector/index.tsx
```diff
@@ -124,7 +126,7 @@ const AppSelector: React.FC = () => {
           )}
         </div>
         <div className="AppSelector__wrapper">
-          <Label color="main">Subscribed</Label>
+          {!empty ? <Label color="main">Subscribed</Label> : null}
```
Cali93, line 129:
```suggestion
          {!empty && <Label color="main">Subscribed</Label>}
```

Cali93 (approved):
LGTM 👍

src/components/notifications/AppSelector/index.tsx
devceline: doesn't this sometimes cause "false" or something to be in the markup

devceline (approved):
LGTM

src/components/notifications/AppSelector/index.tsx
enesozturk: This depends on what the `empty`'s type and value is. Since it's value checking both loading and filteredApp's length, it'll be boolean every time, and `&&` will work as expected. 

Yet, we should still consider following better approaches while doing conditional renderings. Developers shouldn't need to watch the value of the `empty`'s value manually. Instead, we should use the ternary operator. 

Please check this: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

Cali93: Yes I'm aware of that "rule" but it's bloating the code and if the condition is based on a boolean you can always avoid using a ternary in a reliable way.

Up to your preferences I guess.

PR #337, opened 2024-01-05
fix: app selector item image styles

enesozturk:
# Description

A slight update to the image styles of the app selector item.

# Type of change

Slack conversation: https://walletconnect.slack.com/archives/C044SKFKELR/p1704384062573449

### DIFF
--- a/src/components/notifications/AppSelector/AppSelector.scss
+++ b/src/components/notifications/AppSelector/AppSelector.scss
@@ -280,13 +280,17 @@
       &__image {
         width: 100%;
         height: 100%;
-        border-radius: 2em;
+        border-radius: 1em;
       }
 
       @media only screen and (max-width: 700px) {
         border-radius: 1.5em;
         width: 3rem;
         height: 3rem;
+
+        &__image {
+          border-radius: 1.5em;
+        }
 
         &::after {
           border-radius: 1.5rem;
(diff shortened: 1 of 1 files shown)


index.html
devceline, line 1: Why change this file here for app selector styles?

devceline (commented):
Looks like some changes from the PWA branch are here

enesozturk:
@devceline should be fixed now, since on the mobile, the images have different sizes, previous value doesn't meet requirements for fully rounded image

chris13524:
I think still broken?

image

enesozturk:
@chris13524 Did you try with the recent preview link? https://app-web3inbox-r0a1vn87u-walletconnect1.vercel.app/

Cannot reproduce it with my device (iOS 17), I'm thinking if `rem` is not responding as expected in some devices, I'll need to check in some simulators, don't have Android device

chris13524:
@enesozturk the current build looks good.

I don't understand why we have multiple border radiuses though, we should just have 1 and then e.g. `overflow: hidden;` and clip the content

chris13524 (approved):

enesozturk:
@chris13524 

We use `box-shadow` instead of `border` to give inset (inside) borders since CSS doesn't have inset border feature. 

Another problem is that you cannot use inside shadows for an element that renders the image inside. So the `overflow` approach doesn't work here. 

```jsx
<div> /* Using inset shadow doesn't work since the image overrides it. */
 <img/>
</div>
```

Instead, we're using `::after` CSS selector to be able to set `box-shadow` over an image.

Issue #341, opened 2024-01-05
Dev mode duplicates projects on discover page


chris13524:
Enter `gm.walletconnect.com` on the discover view, you'll get a lot of duplicates. [Video](https://walletconnect.slack.com/archives/C044SKFKELR/p1704468389769569)

PR #342, opened 2024-01-05
fix: projects duplicate when have a dev mode

enesozturk:
# Description

Fixes a logic error on the `useNotifyProjects` hook when have a dev mode enabled

# Type of change

# Checklist:

Closes https://github.com/WalletConnect/web3inbox/issues/341

### DIFF
--- a/src/utils/hooks/useNotifyProjects.ts
+++ b/src/utils/hooks/useNotifyProjects.ts
@@ -34,10 +34,10 @@ const useNotifyProjects = () => {
       if (filterAppDomain) {
         explorerUrlAppDomain.searchParams.set('appDomain', filterAppDomain)
         const domainProjectsResponse = await fetch(explorerUrlAppDomain)
-	if(domainProjectsResponse.ok) {
+        if (domainProjectsResponse.ok) {
           const domainProjectsData = await domainProjectsResponse.json()
-          domainProjects = [...discoverProjects, domainProjectsData.data] as INotifyProjectWithComingSoon[]
-	}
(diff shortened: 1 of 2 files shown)


chris13524 (approved):
Did not test

enesozturk:
> Did not test

@chris13524 means ok?

enesozturk:
Ah ok, it's ok as I tested

chris13524 (changes requested):
Nope still something broken. Used `gm.walletconnect.com` and I see 2 of them when there should just be 1.

image

chris13524:
@enesozturk `gm.walletconnect.com` showing as dev mode even though it's featured:

image

enesozturk:
@chris13524 I think we should have talked about this in more detail. What I thought is that it should show the dev badge if I put `gm.walletconnect.com` since I enabled dev mode 'intentionally', so it's an indicator the domain has been set in the settings. Doesn't it make sense?

chris13524 (approved):
@enesozturk I recall we did verbally talk about it and we had said on not displaying as dev mode once the project becomes approved/featured. But we didn't write it down or anything in a table which would've been useful. I'm happy with this though if you think it's better

enesozturk:
@chris13524 If you already talked about it, it makes sense. 

The functionality of the discovery page is a bit weird tho. For non-featured apps, we still show a dev badge but it's actually a page to discover apps. It feels better to list my development apps on another page like "My apps". I'm aware it might not be ideal at this moment but we can consider improving this UX imo

PR #349, opened 2024-01-08
Add history support via new history package

devceline:
# Description
Add history support by adding infinite scrolling using the new notify client fetching.

## Packages Bumps
- Bump notify client to `0.16.0`
## Registration Updates
- Update `register` methods due to breaking changes
- Did not properly refactor to take advantage of the method separation, but this works. I do not want to change registration logic at this crucial time.
## Notification Fetching
- Added reducer to handle fetching message history
- Added infinite scrolling logic on top of said reducer to keep track of messages
- Use this in `AppNotifications` 
- Dedup logic is done using a `Set` (`existingIds`)
- Notifications are kept track of in an array outside Notify client `fullNotification`

# Type of change

# Checklist:

# Notes and Questions

### Why isn't this functionality in `notify-client-js`
Notify client can not make assumptions about how notifications will load. Could be paginated, could be infinite scroll etc. 

### Why is this in a reducer?
It was the only way I could get the `useCallback` to not constantly refresh since we depend on a `set` and and `array` to update the data properly


package.json
```diff
@@ -16,9 +16,9 @@
   },
   "dependencies": {
     "@sentry/react": "^7.64.0",
-    "@walletconnect/core": "^2.10.5",
-    "@walletconnect/identity-keys": "^0.2.2",
-    "@walletconnect/notify-client": "^0.14.5",
+    "@walletconnect/core": "2.11.0",
+    "@walletconnect/identity-keys": "^1.0.1",
+    "@walletconnect/notify-client": "^0.16.0-5ff9624",
```
bkrem, line 21: Need to bump to 0.16.0 exact I believe (saw you merged the release PR for notify-client)

```suggestion
    "@walletconnect/notify-client": "0.16.0",
```

src/w3iProxy/notifyProviders/externalNotifyProvider.ts
```diff
@@ -74,14 +74,9 @@ export default class ExternalNotifyProvider implements W3iNotifyProvider {
...
-  public async register(params: { account: string; domain: string; isLimited?: boolean }) {
-    return this.postToExternalProvider('register', {
-      account: params.account,
-      isLimited: params.isLimited,
-      domain: params.domain,
-      // Signing will be handled wallet-side.
-      onSign: async () => Promise.resolve('')
-    })
+  public async register() {
+    // TODO: remove this whole provider
```
bkrem, line 78: Can we just remove `externalNotifyProvider` if it's now unused instead of setting a TODO or are there concerns around doing this?

src/reducers/notifications.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export type NotificationsActions = {
+  type: 'FETCH_NOTIFICATIONS'
+  notifications: NotifyClientTypes.NotifyMessage[]
+  topic: string
+  hasMore: boolean
+}
+
+// Opted for a reducer since the state changes are complex enough to warrant
+// changes to a set and an array and having all that inside the hooks would
```
bkrem, line 21: Readability. 

Fully agree on opting for a reducer pattern here btw, if anything generally underutilised these days with the over-emphasis on `useState + useEffect`

```suggestion
// changes to a set and an array. Having all that inside the hooks would
```

src/reducers/notifications.ts
```diff
@@ -0,0 +1,52 @@
...
+  state: NotificationsState,
+  action: NotificationsActions
+): NotificationsState => {
+  const topicState = state[action.topic] as TopicNotificationsState | undefined
+
+  switch (action.type) {
+    case 'FETCH_NOTIFICATIONS':
+      const ids = topicState?.existingIds || new Set<string>()
+      const filteredVals = action.notifications.filter(val => !ids.has(val.id))
+      const uniqueIds = action.notifications.map(notification => notification.id)
```
bkrem, line 33: Aren't these all incoming ids? What makes them unique at this stage before being added to a `Set`?

src/reducers/notifications.ts
```diff
@@ -0,0 +1,52 @@
...
+
+      const fullNotifications = topicState?.fullNotifications || []
+      const newFullIdsSet = new Set(topicState?.existingIds || [])
+
+      for (const val of uniqueIds) {
+        newFullIdsSet.add(val)
```
bkrem, line 39: I think you should be able to simplify this whole section into one `Set` assignment unless I'm missing a step:

```
const newFullIdsSet = new Set([...array1, ...array2]);
```

src/reducers/notifications.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+
+      return {
+        ...state,
+        [action.topic]: {
+          ...topicState,
+          existingIds: newFullIdsSet,
+          fullNotifications: fullNotifications.concat(filteredVals),
```
bkrem, line 47: Nit: think it's preferable to do this `concat` operation further up before the return for consistency, as with the Ids `Set`

src/w3iProxy/notifyProviders/internalNotifyProvider.ts
```diff
@@ -107,21 +107,36 @@ export default class InternalNotifyProvider implements W3iNotifyProvider {
...
+      allApps: !Boolean(params.isLimited)
+    }
+
+    if (this.notifyClient.isRegistered(props)) {
+      return this.notifyClient.identityKeys.getIdentity({ account: props.account })
+    }
+
+    const preparedRegistration = await this.notifyClient.prepareRegistration(props)
+
+    console.log('Firing signature')
```
bkrem, line 122:
```suggestion

```

bkrem (approved):
🚀 🚀

src/w3iProxy/notifyProviders/externalNotifyProvider.ts
devceline: It would require a lot of code changes and refactors, not the place for it.

src/reducers/notifications.ts
devceline: This would be a bit too slow for my liking, if hypothetically scrolling down a list of a lot of GMs or something. We get minor optimization for no noticable reduction in readability

src/reducers/notifications.ts
devceline: Ah, yes. Correct. Renaming

src/pages/ScanQrCode/ScanQrCode.scss
```diff
@@ -96,7 +96,8 @@
     --corner-width: 10px;
 
     z-index: 2;
-    background: linear-gradient(
+    background:
```
enesozturk, line 99: How did these files changed since we didn't make updates to Prettier?

src/w3iProxy/notifyProviders/internalNotifyProvider.ts
```diff
@@ -107,21 +107,35 @@ export default class InternalNotifyProvider implements W3iNotifyProvider {
...
+    const preparedRegistration = await this.notifyClient.prepareRegistration(props)
+
+
+    const signature = await (async message => {
+      this.emitter.emit('notify_signature_requested', { message })
+      return new Promise<string>(resolve => {
+        this.emitter.on(
+          'notify_signature_delivered',
+          ({ signature: deliveredSyncSignature }: { signature: string }) => {
+            console.log('getting...')
```
enesozturk, line 129: Unless it's not a helpful for debugging, we could remove

enesozturk (approved):
LGTM 🚀

src/App.scss
```diff
@@ -40,7 +40,10 @@
     height: auto;
     min-height: 100svh;
     overflow-x: hidden;
-    padding-bottom: 90px;
+
+    @media only screen and (max-width: 700px) {
+      padding-bottom: calc(env(safe-area-inset-bottom) + 48px + 2rem);
```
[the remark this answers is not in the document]
devceline: Please create an issue for it so we don't forget

PR #352, opened 2024-01-08
feat: add intersection observer and refactor layout styles

enesozturk:
- add intersection observer for the notification history
- fix layout styling issues
- fix mobile safe area view gap issues


src/components/layout/MobileFooter/MobileFooter.scss
```diff
@@ -9,11 +9,20 @@
...
+    display: flex;
+    align-items: center;
+    padding: 0.5rem 2rem;
+
     svg {
-      width: 1.5rem;
+      width: 24px;
+      min-width: 24px;
+      height: 24px;
+      min-height: 24px;
```
devceline, line 25: why switch to `px` units here?

devceline (approved):
LGTM, ya I need to look up this observer thing. My old scroll function is not as smooth as this. LFG :ship:

src/components/layout/MobileFooter/MobileFooter.scss
enesozturk: Oh, had a really weird issue on the mobile footer; when we switch between tabs, the icons are shifting. I was trying to find the cause, and needed to make sure their sizing, I think it's not what I thought, so I'll be reverting those, good catch

src/utils/hooks/useNotificationsInfiniteScroll.tsx
```diff
@@ -0,0 +1,79 @@
...
+      const newNotifications = await notifyClientProxy.getNotificationHistory({
+        topic,
+        limit,
+        startingAfter: lastMessageId
+      })
+
+      dispatch({
+        type: 'FETCH_NOTIFICATIONS',
+        notifications: newNotifications.notifications,
+        uniqueIds: newNotifications.notifications.map(notification => notification.id),
```
devceline, line 34: nit (non blocking): no need to pass `uniqueIds` since we can just calculate that in the reducer now

src/reducers/notifications.ts
enesozturk, line 1: I've placed this file in a separate folder because as our app grows, we'll likely add more reducers for various parts of the app. It's best to have them organized in one place from the start.

It’s clearer to keep these files separate from the components folder. This helps in distinguishing between different types of files and their purposes.

Our components, along with libraries, utilities, and reducers, should be easily movable. This means we can change their location without issues. When we refactor or reorganize, our components, we might need to access the reducers from various places in our app. By separating them, we avoid situations where one component unnecessarily depends on or accesses the files of another component.

devceline: Good call

src/utils/hooks/useNotificationsInfiniteScroll.tsx
enesozturk: Exactly!

devceline (approved):
:ship: merge it to my branch please and let's get this ball rolling

PR #354, opened 2024-01-09
refactor: update app card styles

enesozturk:
# Description

The current styling solution for the app cards causing some performance issues. The `filter` or `backdrop filter` is one of the most expensive stylings and when we put more value on the `blur` filter. It'll dramatically drop the performance. We can see the effect of this in the Discover page on Safari. While we scroll, the apps are rendered to the screen with a lag.

This PR includes a refactor to optimize the background image styles of the App card. 

# Type of change

# Checklist:

Closes https://github.com/WalletConnect/web3inbox/issues/355

### DIFF
--- a/src/components/notifications/AppExplorer/AppCard/AppCard.scss
+++ b/src/components/notifications/AppExplorer/AppCard/AppCard.scss
@@ -29,13 +29,12 @@
      */
     z-index: 0;
     position: absolute;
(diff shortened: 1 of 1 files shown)


chadyj:
Tested and it works!

Please also test scrolling with lots of notifications.

devceline (approved):
LGTM

PR #358, opened 2024-01-09
fix: use interval to load state incrementally

### LINKED ISSUE #326: Not all subscriptions showed up on initial load
**Describe the bug**
Not all subscriptions showed up on initial load.
only 1 out of 3 which were subscribed in-app)

**To Reproduce**
1. Visit [app.web3inbox.com](http://app.web3inbox.com/)
2. Wipe out all localStorage and sessionStorage
3. Empty cache and hard reload
4. Enter Vercel password
5. Connect Wallet
6. Select WalletConnect
7. Sign message
8. Allow Notifications
9. Expect Subscribed apps from apps that I’ve used
   1. This was the first bug because I saw one app
   2. But I had more subscribed ones and they were missing
10. Subscribe to PancakeSwap
    1. Subscribed successfully to PancakwSwap
    2. suddenly my other missing subscriptions showed up
11. Unsubscribe all of them
12. Empty cache and hard reload
13. Enter Vercel password
14. Connect Wallet
15. Select Metamask
16. Sign Message
17. No subscribed apps
    1. This was working fine so no complains here
    2.
18. Subscribe to GM app
    1. This had a spinner on the button itself but it ended too quick
    2. The spinner stopped but the app didn’t show as subscribed
    3. A bit later the app shows as subscribed
19. Subscribe to PancakeSwap
    1. Same issue as before where the spinner stopped too quickly
    2. An error message showed “subscribe failed”
    3. Later the app showed as subscribed
20. Did a soft refresh of the page
21. The Wallet was still connected as expected
    1. However I noticed that I can see the discovery where apps show as unsubscribed
    2. I would expect here for the Discovery view to have a loading state
22. Tried the settings
    1. Added “display app on discover page”
    2. Typed “robots.farm”
    3. There was no feedback or toast when typing it
    4. This filtered out the other apps on Discovery
    5. Only saw Robots.farm on the Discovery page but expected it to be additive
    6. Then tried typing only “robots.far”
    7. Expected error message or any negative feedback but notihing happened
    8. Visited the discovery page and it was empty
    
**Expected behavior**
All subscriptions load

**Screenshots**

**Additional context**
n/a

devceline:
# Description

- changed the `setTimeout` to `setInterval` in the initial loading for `notifyHooks` state
- Show loading indicator while the interval is running
# Type of change

# Fixes/Resolves (Optional)

Resolves #326 

# Checklist:

### DIFF
--- a/src/components/notifications/AppSelector/index.tsx
+++ b/src/components/notifications/AppSelector/index.tsx
@@ -168,8 +168,7 @@ const AppSelector: React.FC = () => {
                   </AnimatePresence>
                 )
               })}
-            {loading || !subscriptionsFinishedLoading
-              ? Array(3)
+            {loading || !subscriptionsFinishedLoading? Array(3)
                   .fill(<LinkItemSkeleton />)
                   .map(x => x)
               : null}
--- a/src/components/notifications/AppSelector/index.tsx
+++ b/src/components/notifications/AppSelector/index.tsx
@@ -32,7 +32,7 @@ const AppSelector: React.FC = () => {
   const isMobile = useIsMobile()
   const [filteredApps, setFilteredApps] = useState<NotifyClientTypes.NotifySubscription[]>([])
   const [loading, setLoading] = useState(true)
-  const { activeSubscriptions } = useContext(W3iContext)
+  const { activeSubscriptions, watchSubscriptionsComplete: subscriptionsFinishedLoading } = useContext(W3iContext)
   const { projects } = useNotifyProjects()
 
   const empty = !loading && filteredApps.length === 0
@@ -131,11 +131,6 @@ const AppSelector: React.FC = () => {
         <div className="AppSelector__wrapper">
           {!empty ? <Label color="main">Subscribed</Label> : null}
           <ul className="AppSelector__list">
-            {loading
-              ? Array(3)
-                  .fill(<LinkItemSkeleton />)
-                  .map(x => x)
-              : null}
             {!loading &&
               filteredApps?.map(app => {
                 return (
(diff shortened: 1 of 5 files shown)


src/components/notifications/AppSelector/index.tsx
```diff
@@ -173,6 +168,10 @@ const AppSelector: React.FC = () => {
                   </AnimatePresence>
                 )
               })}
+            {loading || !subscriptionsFinishedLoading? Array(3)
```
enesozturk, line 171: Not urgent but we should create the array on the top of the component so it doesn't have to be recreated on every render: 

```jsx
const LoadingItems = Array(3).Fill(<LinkItemSkeleton />).map(x => x)
```

src/contexts/W3iContext/hooks/notifyHooks.ts
```diff
@@ -44,12 +46,18 @@ export const useNotifyState = (w3iProxy: Web3InboxProxy, proxyReady: boolean) =>
...
+  // it takes time for handshake (watch subscriptions) to complete
+  // load in progress state using interval until it is
   useEffect(() => {
+    if(watchSubscriptionsComplete) {
+      return noop;
+    }
+
     // Account for sync init
-    const timeoutId = setTimeout(() => refreshNotifyState(), 100)
+    const timeoutId = setInterval(() => refreshNotifyState(), 500)
```
enesozturk, line 57: Depending on the cost of the fetching notify state on the server, I think this could be even `1000` or `2000`,  wdyt?

enesozturk (approved):
Lets goo

src/contexts/W3iContext/hooks/notifyHooks.ts
devceline: This does not fetch from notify server, just locally

src/components/notifications/AppSelector/index.tsx
devceline: Yes, let's handle this in a different PR. I really want to handle everything a bit faster

Issue #329, opened 2024-01-04
Refreshing the Discovery page doesn’t have a loading state so apps show unsubscribed


chadyj:
**Describe the bug**
Refreshing the Discovery page doesn’t have a loading state so apps show unsubscribed

**To Reproduce**
1. Subscribe to some apps
2. Go to discover page
3. Refresh the page

**Expected behavior**
The subscription button state shows correct status

**Actual behavior**
Subscirption state button changes after load

**Screenshots**
n/a

**Desktop (please complete the following information):**
n/a

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

enesozturk:
We recently made improvements and due to lack of ability to handle the loading state for all subscriptions at once, we're going to show loading skeletons until the notify server handshake is done. 

So this should be closed once we merge https://github.com/WalletConnect/web3inbox/pull/362

PR #362, opened 2024-01-10
refactor: revert the refresh notify state interval logic

enesozturk:
# Description

The code piece in [this useEffect](https://github.com/WalletConnect/web3inbox/blob/main/src/contexts/W3iContext/hooks/notifyHooks.ts#L54-L57)  is running before the notify handshake is done, this causes the same issue we have that the loading skeletons are unmounted before the handshake completed and as a result, creates the same bad UX issue.

# Type of change

# Checklist:

# Additional notes

Closes https://github.com/WalletConnect/web3inbox/issues/329


devceline (commented):
This is causing permanent skeletons sometimes, can you double check that issue?

PR #367, opened 2024-01-10
chore: update condition

devceline:
# changes
- use `isInitialSubscriptionLoadComplete` for more accurate status

### DIFF
--- a/src/w3iProxy/notifyProviders/internalNotifyProvider.ts
+++ b/src/w3iProxy/notifyProviders/internalNotifyProvider.ts
@@ -18,8 +18,6 @@ export default class InternalNotifyProvider implements W3iNotifyProvider {
   public providerName = 'InternalNotifyProvider'
   private readonly methodsListenedTo = ['notify_signature_delivered']
 
-  private gotFirstWatchSubscriptionsResponse = false;
-
   public constructor(emitter: EventEmitter, _name = 'internal') {
     this.emitter = emitter
   }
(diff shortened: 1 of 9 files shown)


src/w3iProxy/notifyProviders/internalNotifyProvider.ts
```diff
@@ -106,8 +103,8 @@ export default class InternalNotifyProvider implements W3iNotifyProvider {
 
   // ------------------- Method-forwarding for NotifyClient -------------------
 
-  public isInitialSubscriptionLoadComplete() {
-    return this.gotFirstWatchSubscriptionsResponse;
+  public hasFinishedInitialLoad() {
+    return this.notifyClient?.hasFinishedInitialLoad() || false;
```
enesozturk, line 107: Do we need `|| false` here? If yes, do we have a case that `this.notifyClient?.hasFinishedInitialLoad()` always falsy?

devceline: just added the `|| false` instead `notifyClient` is `undefined`

enesozturk: @devceline I mean that on the Notify side, if somehow it's not set to true, we'll still show skeletons infinitely, right?

devceline: Yes, but that should never ever happen. We set this to true even if there an error

enesozturk: Understood. We should always handle all cases. Wanted to double check

enesozturk (approved):

Issue #348, opened 2024-01-08
Discover ordering incorrect on mobile


chadyj:
Mobile has different ordering than desktop.

Eg GM is below Gempad (which should be last since it is coming soon), and SPACE ID

![Image](https://github.com/WalletConnect/web3inbox/assets/116099/c0e5fa17-9253-4f01-a9ac-51d4ba03e33d)

PR #391, opened 2024-01-13
refactor: update breakpoints, handle breakpoint for explorer columns

enesozturk:
# Description

Refactored the explorer page to handle window resizing and render the apps in the right order with conditionally rendered columns. Separated some app explorer components (skeletons and app listing components) for better maintainability and readability.

ℹ️ We should avoid JS solutions for layout stylings. For both performance and hydration issues, it's a bad practice. Since we're using a masonry layout and it's not natively supported by CSS, we need to use JS.  

Along with the layout changes, refactored the screen breakpoint sizes. Since the values like `768px`, and `1024px` are common values to separate different devices (tablets, laptops, etc.) it's better to use them while doing media queries. These will provide more real-life use cases.

# Type of change

# Fixes/Resolves (Optional)

Closes https://github.com/WalletConnect/web3inbox/issues/348

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### DIFF
--- a/src/components/notifications/AppExplorer/AppExplorer.scss
+++ b/src/components/notifications/AppExplorer/AppExplorer.scss
@@ -4,6 +4,7 @@
   flex-direction: column;
   align-items: center;
   max-width: 100vw;
+  padding: 0 1.5rem;
 
   backface-visibility: hidden;
   -moz-backface-visibility: hidden;
@@ -13,6 +14,7 @@
 
   @media only screen and (max-width: 1300px) {
     margin-right: 0%;
+    padding: 0 1rem;
   }
 
   @media only screen and (min-width: 768px) {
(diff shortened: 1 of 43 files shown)


chadyj:
Tested on safari mac. Vertical line appeared

Screenshot 2024-01-13 at 14 56 07

chadyj:
The ordering is fixed though :+1:

devceline (approved):
LGTM

src/components/notifications/AppExplorer/AppExplorer.scss
```diff
@@ -85,23 +85,24 @@
 
   &__apps {
     width: 100%;
-    max-width: 700px;
+    max-width: 768px;
```
enesozturk, line 88: This shouldn't screen width. This container intentionaly set to 700px by the designer. It's not a screen breakpoint.

enesozturk (commented):
Reviewed

enesozturk:
Tested on various devices

PR #392, opened 2024-01-13
refactor: move the subscription complete handler on useEffect

enesozturk:
# Description

Since we're calling the `handleSubmit` at once and the `subscribed` and `subscribing` values changing over time. We should watch these values in `useEffect` to handle subscribed state and take action (like show a toast message).

# Type of change

# Fixes/Resolves (Optional)

Closes https://github.com/WalletConnect/web3inbox-planning/issues/51

https://github.com/WalletConnect/web3inbox/assets/19428358/71063d75-68bb-4c99-adf7-909c5b8e71ab

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### DIFF
--- a/src/components/notifications/AppExplorer/AppCard/index.tsx
+++ b/src/components/notifications/AppExplorer/AppCard/index.tsx
@@ -62,12 +62,6 @@ const AppCard: React.FC<AppCardProps> = ({
         return
       }
 
-      if (subscribing && subscribed) {
-        showSuccessMessageToast(`Subscribed to ${name}`)
-
-        return
-      }
-
       setSubscribing(true)
       try {
         await notifyClientProxy?.subscribe({
@@ -82,6 +76,12 @@ const AppCard: React.FC<AppCardProps> = ({
     },
     [userPubkey, name, description, logo, url, setSubscribing, subscribed, notifyClientProxy]
   )
+
+  useEffect(() => {
+    if (subscribing && subscribed) {
+      showSuccessMessageToast(`Subscribed to ${name}`)
+    }
+  }, [subscribing, subscribed])
 
   const handleNavigateApp = () => {
     if (subscribed) {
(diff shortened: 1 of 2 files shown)


src/components/notifications/AppExplorer/AppCard/index.tsx
```diff
@@ -83,6 +77,12 @@ const AppCard: React.FC<AppCardProps> = ({
     [userPubkey, name, description, logo, url, setSubscribing, subscribed, notifyClientProxy]
   )
 
+  useEffect(() => {
+    if (subscribing && subscribed) {
+      showSuccessMessageToast(`Subscribed to ${name}`)
```
enesozturk: @devceline Should we call the `setSubscribing(false)` in this block as well? Seems like we don't set it to false if not error occurs.

src/components/notifications/AppExplorer/AppCard/index.tsx
```diff
@@ -99,6 +93,12 @@ const AppCard: React.FC<AppCardProps> = ({
     }
   }
 
+  useEffect(() => {
+    if (subscribing && subscribed) {
+      showSuccessMessageToast(`Subscribed to ${name}`)
```
enesozturk, line 98: @devceline Should we call the `setSubscribing(false)` in this block as well? Seems like we don't set it to `false` if no error occurs.

devceline: I don't think it matters because `subscribed` would be true

devceline (approved):

src/components/notifications/AppExplorer/AppCard/index.tsx
enesozturk: Lets double make sure it's not gonna retriggered

enesozturk (commented):
Reviewed

PR #382, opened 2024-01-11
fix: ensure every error toast logs the underlying error to console

bkrem:
# Description

- Context: https://walletconnect.slack.com/archives/C044SKFKELR/p1704993119733469?thread_ts=1704976288.139559&cid=C044SKFKELR
- We were catching some errors and showing user friendly toasts without processing the underlying error in an observable way.
- Ensure every `catch + showErrorMessageToast` is paired to a `console.error` call
- Any other changes are ESLint/Prettier auto-fixes

# Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Draft PR (breaking/non-breaking change which needs more work for having a proper functionality _[Mark this PR as ready to review only when completely ready]_)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

PR #400, opened 2024-01-15
chore(config): ensure Sentry captures `console.error` calls

bkrem:
# Description

- Follow-up to #382 
- Context: Sentry will _not_ capture explicit `console.error` calls by default unless specifically configured to do so.
- Adds `CaptureConsole` integration with only `error` log level being captured. Context: https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/captureconsole/
- Additionally: updates `@sentry/react` to latest
- Tested that Sentry still triggers as expected on the preview

## Alternatives Considered

- Rewiring all the `catch` statements touched in #382 to `throw`. This is risky 1 day before launch bc it will bail on the remaining function and may cause unwanted/unexpected changes to the app's behaviour if not done carefully.
- Adding `Sentry.captureException` alongside the added `console.error()` calls. This will create unnecessary clutter and will leave gaps over time where we forget to add `Sentry.captureException`. Effectively achieving a more cluttered and less reliable version of this top-level configuration.

# Type of change

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "playwright:test": "playwright test"
   },
   "dependencies": {
-    "@sentry/react": "^7.64.0",
+    "@sentry/integrations": "^7.93.0",
+    "@sentry/react": "^7.93.0",
     "@walletconnect/core": "2.11.0",
     "@walletconnect/identity-keys": "^1.0.1",
     "@walletconnect/notify-client": "^0.16.4",
--- a/src/utils/sentry.ts
+++ b/src/utils/sentry.ts
@@ -1,3 +1,4 @@
+import { CaptureConsole } from '@sentry/integrations'
 import * as Sentry from '@sentry/react'
 
 export const initSentry = () => {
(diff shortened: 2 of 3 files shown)


devceline (approved):
LGTM - I think this will add a lot of error logs that are already handled because off the top of my head I believe we `console.error` stuff that's handled. 

This is good as a solution for near launch but I think the alternatives are better long term solutions as this will likely cause unnecessary spam on our sentry (false positives).

So - for now let's merge this to catch all errors. However, after launch let's look at your alternatives with the throws.

bkrem:
> This is good as a solution for near launch but I think the alternatives are better long term solutions as this will likely cause unnecessary spam on our sentry (false positives).

Completely agree. I will help you triage the Sentry this week btw, but I'd rather we have too much coming in than missing something important that users are seeing.

Long term I think:
* move to simply `throw` to rethrow an error in a controlled way with a useful error message when this is appropriate (i.e. unrecoverable state or we want to bail out of the remaining function)
* Add a self-encapsulated error handler. E.g. in Cloud we always show a toast and then rethrow to top level (hence why this `console.error` detail hasn't been an issue there before)

PR #405, opened 2024-01-16
fix: ensure useEffect does not spam `closeWeb3Modal` + close event

bkrem:
# Description

- Bug: This `useEffect` is called in a loop for the duration of the signature modal being open
	- It was also calling `closeWeb3Modal` unchecked i.e. even if w3m was already in a closed state
	- This was causing our analytics endpoint to be spammed with `MODAL_CLOSE` events **which could lead to unintentional request floods from users that leave the app in this state.**
- **This patch ensures we first check `open: boolean` state for w3m before explicitly calling `closeWeb3Modal`**
- **This patch does not root cause why this useEffect goes into a loop once the signature modal opens, this is TBD.**

# Type of change

## Reproduction

1. Go to app.web3inbox.com (or a preview that has w3m analytics enabled)
2. Use a fresh account or reset client state to ensure you get the secondary "Sign notify message" popup after w3m connect
3. Observe network tab after initial w3m connect; spamming the `pulse.walletconnect.com/e` endpoint with `MODAL_CLOSE` events (see screenshot)

## Screenshots

Screenshot 2024-01-16 at 12 10 54

# Checklist:

### DIFF
--- a/src/Modals.tsx
+++ b/src/Modals.tsx
@@ -35,7 +35,8 @@ export const Modals = () => {
   const notificationsEnabled = useNotificationPermissionState()
 
   const notificationModalClosed = checkIfNotificationModalClosed()
-  const explicitlyDeniedOnDesktop = !isMobile() && window.Notification.permission === 'denied'
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  const explicitlyDeniedOnDesktop = !isMobile() && window.Notification?.permission === 'denied'
   const shouldShowChangeBrowserModal = isAppleMobile ? isNonSafari : false
   const shouldShowPWAModal = isMobileButNotInstalledOnHomeScreen && !shouldShowChangeBrowserModal
   const shouldShowSignatureModal = isSignatureModalOpen && !shouldShowChangeBrowserModal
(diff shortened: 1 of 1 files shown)


src/Modals.tsx
```diff
@@ -28,13 +28,14 @@ export const Modals = () => {
...
   const { close: closeWeb3Modal } = useWeb3Modal()
+  const { open: isWeb3ModalOpen } = useWeb3ModalState()
 
   const { notifyRegisterMessage, notifyRegisteredKey, userPubkey } = useContext(W3iContext)
 
   const notificationsEnabled = useNotificationPermissionState()
 
   const notificationModalClosed = checkIfNotificationModalClosed()
-  const explicitlyDeniedOnDesktop = !isMobile() && window.Notification?.permission === 'denied'
+  const explicitlyDeniedOnDesktop = !isMobile() && window.Notification.permission === 'denied'
```
devceline, line 38: Why remove the `?`, access to `Notification` directly breaks the app in certain mobile environments

devceline (commented):
It looks good to me bar the `Notification` access

src/Modals.tsx
bkrem: This was an ESLint autofix I believe, will undo it

enesozturk (approved):

src/Modals.tsx
bkrem: Added it back and removed the ESLint error for that expression so peoples' editors don't auto fix in future.

PR #420, opened 2024-01-26
chore: losslessly compress all img assets

bkrem:
# Description

Vercel bandwidth is our friend :)


enesozturk:
@bkrem @devceline Seems like most of these icons are not used, high possibly they belongs to previous chat application, what about removing them?

If we don't have a plan in the near future to build chat experience, we could remove the chat code along with the assets. That would be better, I could handle this, wdyt?

chris13524:
Should we merge this? Also using Git LFS would be nice.

Re: chat, we don't have plans for the next several months but we may revisit it. I suppose it would be fine to remove it all for now though as we will still have the code in Git history.

enesozturk (approved):

enesozturk:
@bkrem This looks good to me, I'm merging it. Thanks 🙌

@chris13524 totally, I think we should get rid of it until have a stable plan. Too much things to refactor and remove in this codebase. Let's keep it as minimal as possible.

PR #425, opened 2024-01-30
redirect to domain on click

enesozturk:
# Description

On the topic details page header, make the URL clickable.

# Type of change

# Checklist:

[Slack thread](https://walletconnect.slack.com/archives/C044SKFKELR/p1705685614022099)

### DIFF
--- a/src/components/general/Icon/ArrowRightTopIcon.tsx
+++ b/src/components/general/Icon/ArrowRightTopIcon.tsx
@@ -2,35 +2,13 @@ import React from 'react'
 
 const ArrowRightTopIcon: React.FC<React.SVGProps<SVGSVGElement>> = props => {
   return (
-    <svg
-      xmlns="http://www.w3.org/2000/svg"
-      width="16"
-      height="17"
-      viewBox="0 0 16 17"
-      fill="none"
-      {...props}
-    >
-      <g clipPath="url(#clip0_4498_48131)">
-        <path
-          d="M11.3012 5.19971L4.70117 11.7997"
-          stroke="currentColor"
-          strokeWidth="1.25"
-          strokeLinecap="round"
-          strokeLinejoin="round"
-        />
-        <path
-          d="M6 5.19922L11.2993 5.19932V10.5"
-          stroke="currentColor"
-          strokeWidth="1.25"
-          strokeLinecap="round"
-          strokeLinejoin="round"
-        />
-      </g>
-      <defs>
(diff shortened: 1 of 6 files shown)


src/components/notifications/AppNotifications/AppNotificationsHeader/index.tsx
```diff
@@ -25,7 +29,7 @@ const AppNotificationsHeader: React.FC<IAppNotificationsHeaderProps> = ({
 }) => {
   const isMobile = useIsMobile()
   const { dappOrigin } = useContext(W3iContext)
-  const [dropdownToShow, setDropdownToShow] = useState<string | undefined>()
+  const url = new URL('https://' + domain)
```
enesozturk, line 32: Definitely not happy with this. We don't have full URL of the dApp in the metadata from Notify Client. 

@devceline Do you think we could do this better? Maybe like matching, the dApp with projects and finding the project object, and getting the URL? Both looking overengineering but we cannot update the Notify Client for this too.

enesozturk: Let's move the projects to a Context. We can get the project list in the AppCard find the project and use the URL from the project object. That would be a better approach. I wouldn't want to ship this. I'll take a look later

src/components/general/Icon/ArrowRightTopIcon.tsx
devceline, line 1: Is this used anywhere else?

src/components/notifications/AppNotifications/AppNotificationsHeader/index.tsx
devceline: Can we huddle, don't fully understand this here?

devceline (approved):
LGTM

src/components/general/Icon/ArrowRightTopIcon.tsx
enesozturk: Leaving notes as we talked about it with @devceline while pairing:

This is only used a few places and it's already checked if the external link icon is looking good there and all good.

src/components/notifications/AppNotifications/AppNotificationsHeader/index.tsx
enesozturk: Okay let's go with this way since we are already sure that the domain wouldn't be empty as we check. I'd still like to have url from the backend instead of creating it on the UI

Issue #437, opened 2024-02-10
Enable Web3Modal email


chris13524:
Blocked by https://github.com/WalletConnect/web3inbox/issues/436

https://docs.walletconnect.com/web3modal/react/email/

Issue #436, opened 2024-02-10
Update to WAGMI v2


chris13524:
Unblocks https://github.com/WalletConnect/web3inbox/issues/437

https://wagmi.sh/react/guides/migrate-from-v1-to-v2

PR #438, opened 2024-02-10
feat: initialize w3m email alpha

enesozturk:
# Description

Add the email feature of the Web3Modal with the v4 alpha version. This requires a Wagmi v2 upgrade. This PR includes the required changes of the wagmi/view functions/hooks/methods.

# Type of change

# Checklist:

Closes https://github.com/WalletConnect/web3inbox/issues/436
Closes https://github.com/WalletConnect/web3inbox/issues/437

[Slack conversation](https://walletconnect.slack.com/archives/C044SKFKELR/p1707527886425219)

### DIFF
--- a/src/Modals.tsx
+++ b/src/Modals.tsx
@@ -59,7 +59,7 @@ export const Modals = () => {
     if (userPubkey && notifySignatureRequired) {
       if (isWeb3ModalOpen) {
         // Close web3modal in case user is switching accounts
-        closeWeb3Modal()
+        // closeWeb3Modal()
       }
       signatureModalService.openModal()
     } else {
@@ -91,9 +91,7 @@ export const Modals = () => {
 
       {shouldShowPreferencesModalOpen && <PreferencesModal />}
 
-      {shouldShowSignatureModal && (
-        <SignatureModal message={notifyRegisterMessage ?? ''} sender={'notify'} />
-      )}
+      {shouldShowSignatureModal && <SignatureModal message={notifyRegisterMessage ?? ''} />}
 
       {shouldShowPWAModal && <PwaModal />}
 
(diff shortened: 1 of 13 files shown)


arein (approved):
LGTM except the Magic issue that needs to be addressed - maybe zIndex?

src/main.tsx
```diff
@@ -31,7 +31,13 @@ createWeb3Modal({
   themeMode: 'light',
   themeVariables: { '--w3m-z-index': 9999 },
   termsConditionsUrl: TERMS_OF_SERVICE_URL,
-  privacyPolicyUrl: PRIVACY_POLICY_URL
+  privacyPolicyUrl: PRIVACY_POLICY_URL,
+  metadata: {
+    name: 'Web3Inbox',
+    description: 'Web3Inbox App',
+    url: 'https://web3inbox.com',
+    icons: ['https://avatars.githubusercontent.com/u/37784886']
```
chris13524, line 39: Shouldn't this be Web3Inbox logo not WalletConnect?

src/Modals.tsx
```diff
@@ -59,7 +59,7 @@ export const Modals = () => {
     if (userPubkey && notifySignatureRequired) {
       if (isWeb3ModalOpen) {
         // Close web3modal in case user is switching accounts
-        closeWeb3Modal()
+        // closeWeb3Modal()
```
enesozturk, line 62: ### Why we required to disable this

When signing with email, we handle the signature message in the w3m, not in the wallet popup (like MM, or other wallets). It happens inside an iframe that renders [Magic](https://magic.link/)'s message approves screen. So, auto closing causes a UI regression and immediately closes the signature modal. 

Since we don't have a chance to check if the used method is email or not, we need to disable auto closing w3m. This condition is quite a rare case and if some reason the modal opens, users could close it and continue proceeding.

PR #470, opened 2024-03-06
fix: test env vars

chris13524:
- Missing env vars failed during the tests rather than at startup
- Missing vars from .env.example
- Load env vars from .env (assuming before these were set by hand with `export`)
- Rename `messageType` to `notificationType`


devceline (approved):
LGTM

PR #471, opened 2024-03-06
Fix/add ci lints

chris13524:
Adds eslint and prettier to CI

Supersedes #389


src/reducers/notifications.ts
```diff
@@ -81,7 +82,7 @@ export const notificationsReducer = (
           ...topicState,
           existingIds: newFullIdsSet,
           fullNotifications: unshiftedNotifications,
-          hasMore: topicState?.hasMore || false,
+          hasMore: topicState?.hasMore ?? false,
```
devceline, line 85: Again, `hasMore` is a boolean. I think we should change the `??` rule. In this specific instance it won't matter but a blanket change like this is bad

devceline (commented):
Looks good overall, left few comments. Also this needs a rebase

PR #474, opened 2024-03-06
fix: avoid subscribing to non-test apps, subscribe by name not index

chris13524:
Avoid subscribing to random apps by their position on the discover page. Instead, we should only subscribe to internal ones (i.e. Notify Swift Integration Tests Prod). For now subscribing to GM Dapp for the "2 subscriptions" test case, but we must find a way to avoid polluting GM Dapp metrics with ephemeral wallet addresses.


devceline (approved):
LGTM

PR #519, opened 2024-03-27
fix: Playwright tests

### LINKED ISSUE #514: Canary image failing to build; needs to use yarn not npm
https://github.com/WalletConnect/web3inbox/actions/runs/8422088510/job/23061125765

### LINKED ISSUE #525: NEXT_PUBLIC_LOG_LEVEL
Allow custom builds (e.g. builds for Playwright tests) to enable console trace logging to help debug tests in CI. [Slack conversation](https://walletconnect.slack.com/archives/C03RME3BS9L/p1711561253515899?thread_ts=1711559317.364129&cid=C03RME3BS9L)

chris13524:
Fixes the failing build because of using npm not yarn. Fix failing Playwright tests by testing against the built version of the app to avoid long (15s) startup for first test

https://github.com/WalletConnect/web3inbox/actions/runs/8436003975/job/23102606745

Resolves #514 
Resolves #525


tests/shared/constants/index.ts
```diff
@@ -1,7 +1,5 @@
 import type { SessionParams } from '../types'
 
-// Allow localhost
-export const BASE_URL = process.env['BASE_URL'] || 'http://localhost:5173/'
```
chris13524, line 4: Playwright injects a baseURL from the config; we should be using that.

devceline (commented):
https://github.com/WalletConnect/web3inbox/actions/runs/8545112344/job/23413246643 - pipeline failed after being rerun. It failing at an always existing nav link probably means it's more to do with the test rather than actual functionality

devceline (approved):
Changed signature modal logic a bit so that's more reliable - LGTM

Also added retries because inspecting the tests shows a lot of their failures was the QR code just taking an absurd time to load or failing to load at all

PR #561, opened 2024-05-08
fix: failing cases hang

chris13524:
Seems like when a test fails it hangs the ECS task and doesn't exist. Setting `CI=true` will prevent this. [Slack conversation](https://walletconnect.slack.com/archives/C058RS0MH38/p1715173506598119)


devceline (approved):
LGTM

PR #568, opened 2024-05-13
chore: canary trace uploads

chris13524:
Uploads Playwright traces to S3 (like in Web3Modal canary) to enable debugging of failures and latency regressions.


devceline (approved):
LGTM

PR #570, opened 2024-05-14
chore: UI canary 10 minute timeout

chris13524:
Adds a 10 minute timeout

### DIFF
--- a/Dockerfile.canary
+++ b/Dockerfile.canary
@@ -20,4 +20,4 @@ RUN yarn playwright:install
 ENV CI=true
 
 ENV PRE_BUILD=true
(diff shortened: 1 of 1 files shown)


lukaisailovic (approved):
Why 10m?

nopestack (approved):

chris13524:
> Why 10m? Can we get away with with something else?

@lukaisailovic not sure what you are asking. We could make it longer or a little shorter. Canaries should run within like 3 minutes. Made it 10 to give ample room for slowdowns. The purpose of the timeout is to prevent the task from executing indefinitely if someone breaks the Playwright config.

